PR #3, opened 2017-04-30
Fix travis build and emit canonical name for capitalized proto fields in messages.

### LINKED ISSUE #2: All caps field names don't become camelcase
``` proto
message Example {
    int32 FIELD = 1;
}
```

creates a .d.ts file containing

``` ts
export class Example extends jspb.Message {
  getFIELD(): number;
  setFIELD(value: number): void;

  // ...
}
```

but a .js files containing

``` js
/**
 * optional int32 FIELD = 1;
 * @return {number}
 */
proto.api.Example.prototype.getField = function() {
  return /** @type {number} */ (jspb.Message.getFieldWithDefault(this, 1, 0));
};

/** @param {number} value */
proto.api.Example.prototype.setField = function(value) {
  jspb.Message.setField(this, 1, value);
};
```

Note that these function names are not in caps. When calling `getFIELD()` this will fail silently in Chrome. Instead, ts-protoc-gen should camelcase the getter/setter method names to match js output.

paranoiacblack:
This fixes #2 as well as updating tests so `npm test` passes on travis again.

### DIFF
--- a/src/ts/message.ts
+++ b/src/ts/message.ts
@@ -58,7 +58,7 @@ export function printMessage(fileName: string, exportMap: ExportMap, messageDesc
       }
       existing.push(field);
     }
-    const snakeCaseName = field.getName();
+    const snakeCaseName = field.getName().toLowerCase();
     const camelCaseName = snakeToCamel(snakeCaseName);
     const withUppercase = uppercaseFirst(camelCaseName);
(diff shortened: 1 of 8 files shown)


MarcusLongmuir:
Thanks for the fix!

PR #18, opened 2017-08-08
All fields in proto3 syntax are optional

mackross:


jonny-improbable:
Sorry to close this but I have no context and it's drifted a long way from master. Please could you open a bug if there is an issue here you are trying to resolve? Thanks!

PR #26, opened 2017-09-13
Update package.json

danbopes:
Fix up the repository URL so it shows correctly on npmjs.com


jonnyreeves:
Thanks, fix will be pushed with the next release

PR #29, opened 2018-01-13
Create stubs for handling requests in a slightly easier manner

thakkarparth007:
I have modified the code so that the user can directly use `stub.request(requestObject)` instead of having to call `grpc.unary` or `grpc.invoke` everywhere.

The generated code will allow unary-unary calls to be made as such:

```javascript
async function login() {
    const loginRequest = new LoginRequest();
    loginRequest.setEmailId("email@gmail.com");
    loginRequest.setPassword("password");

    try {
        const resp = await stub.login(loginRequest, optionalMetadata);
        // resp is an instance of LoginResponse
    } catch (e) {
        // log the error
    }
}
```

And for server-streaming functions, the user can use the following syntax:
```javascript
async function joinNotificationsStream() {
    const getNotificationsRequest = new GetNotificationsRequest();
    const stream = stub.getNotifications(getNotificationsRequest, optionalMetadata);

    for await (const notif of stream) {
        console.log(notif); // notif will be an instance of the response type of GetNotifications;
    }
}
```

Since the generated code is heavy with nested indentation, I have made a couple helper methods to write code in a pattern that matches the expected indentation. However I'm open to following any alternative convention.

Also, the current code isn't meant to handle all use cases (like looking at headers and trailers in the response). Currently it's meant to be used for my own purposes, but I'd be happy to generalize it a bit more as per your suggestions.

There's one line in the unary-unary stub that looks at the grpc status code as well as the status code embedded in the response object (I have that in my project), and the stub throws an error if either of them is non-OK code. I've left it as it is right now to show that in case the application level has its own error handling on top of grpc, having errors thrown from the stub can be quite convenient. Again, this makes sense in my own project, but might not in a general library like grpc-web. I can remove it if required.


easyCZ:
That's pretty cool! Will look through the code and have a think how we can incorporate this. Thanks for raising

thakkarparth007:
@easyCZ, any updates?

jonny-improbable:
Closed in favour of #44

PR #37, opened 2018-02-22
Add path to service definition

josephn123:
Adds path to generated service definition.

This enables the generated service definition (NAME_pb_service.ts) to be used directly with grpc.Server::addService(), after extending the generated class with the appropriate serialiser and deserialiser.

Most importantly, it ensures the knowledge of the path is embedded within the service definition for use by client and server without having to specify it in two places.

### DIFF
--- a/src/ts/fileDescriptorTSServices.ts
+++ b/src/ts/fileDescriptorTSServices.ts
@@ -5,9 +5,6 @@ import {FileDescriptorProto} from "google-protobuf/google/protobuf/descriptor_pb
 import {WellKnownTypesMap} from "../WellKnown";
 import {getFieldType, MESSAGE_TYPE} from "./FieldTypes";
 
-const makeSerializer = (messageType: string) => `(obj: ${messageType}) => new Buffer(obj.serializeBinary())`;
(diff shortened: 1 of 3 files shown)


mainnika:
@easyCZ with these changes it is possible to use ts-protoc-gen tool directly with [grpc-node](https://github.com/grpc/grpc-node) library and make a node grpc server or client with typescript.

E.g.:
```
import { Server } from 'grpc';
import * as yourSvc from './your_pb_service';
import * as your from './your_pb';

const server = new Server();
server.addService(yourSvc.Service, {
	Func1: (foo, callback) => callback(null, new your.Response()),
});
```


But @josephn123, why did you remove serialisers?

josephn123:
@mainnika I removed the serialisers because Buffer is node-specific. The declarations would be made incompatible with any non-node (e.g. client-side) grpc typescript clients.

@easyCZ The path is derivable, however grpc-requires the path property as part of the declaration. I don't think it's wise to require clients to have to derive this path property themselves by amalgamating the other properties, all "knowledge" should be embedded in the declaration, especially for something as important as the path.

jonny-improbable:
Closing this as stale given the recent changes that have landed in master; seeing as this is a grpc-node specific feature you may wish to consider incorporating your changes into #42?

Issue #25, opened 2017-09-08
 The embedded message caused the TS6133 error


linlexing:
The embedded message incorrectly generates the import statement.

such as:

inner_data.proto:
```protobuf
syntax = "proto3";
package test;
message innerData{
    string name=1;
}
``` 
service_test.proto:
```protobuf
syntax = "proto3";
import "inner_data.proto";
package test;
message RequestMessage{
    innerData data=1;
}
message ResultMessage{
    string data=1;
}
service RequestService {
    rpc GetRequest(RequestMessage) returns (ResultMessage);
}
```

Generated file:
service_test_pb_service.ts:
```typescript
// package: test
// file: service_test.proto

import * as service_test_pb from "./service_test_pb";
import * as inner_data_pb from "./inner_data_pb";//<<----Cause TS6133 'declared but never used' error
export class RequestService {
  static serviceName = "test.RequestService";
}
export namespace RequestService {
  export class GetRequest {
    static readonly methodName = "GetRequest";
    static readonly service = RequestService;
    static readonly requestStream = false;
    static readonly responseStream = false;
    static readonly requestType = service_test_pb.RequestMessage;
    static readonly responseType = service_test_pb.ResultMessage;
  }
}
```

MarcusLongmuir:
@linlexing: The cause of this is that the service generation performed by [fileDescriptorTSServices.ts](https://github.com/improbable-eng/ts-protoc-gen/blob/master/src/ts/fileDescriptorTSServices.ts) doesn't check which namespaces (e.g. `inner_data_pb`) are actually going to be printed in the declaration before printing them as imports.

The fix would be to first pass over all of the message definitions and only print the imports that are necessary.

jonny-improbable:
This appears to have been fixed in #38; please re-open if that's not the case :)

PR #38, opened 2018-02-24
Don't output variables that are not used in typescript service definition

jonbretman:
I got burnt by https://github.com/improbable-eng/ts-protoc-gen/issues/25 too so thought I'd have a go at fixing it.

Will also have a go at adding some tests for this.

@linlexing @MarcusLongmuir

### DIFF
--- a/src/ts/fileDescriptorTSServices.ts
+++ b/src/ts/fileDescriptorTSServices.ts
@@ -5,6 +5,20 @@ import {FileDescriptorProto} from "google-protobuf/google/protobuf/descriptor_pb
 import {WellKnownTypesMap} from "../WellKnown";
 import {getFieldType, MESSAGE_TYPE} from "./FieldTypes";
 
+function isUsed(fileDescriptor: FileDescriptorProto, pseudoNamespace: string, exportMap: ExportMap) {
+  return fileDescriptor.getServiceList().some(service => {
+    return service.getMethodList().some(method => {
+      const requestMessageTypeName = getFieldType(MESSAGE_TYPE, method.getInputType().slice(1), "", exportMap);
+      const responseMessageTypeName = getFieldType(MESSAGE_TYPE, method.getOutputType().slice(1), "", exportMap);
+      const namespacePackage = pseudoNamespace + ".";
+      return (
+        requestMessageTypeName.indexOf(namespacePackage) === 0 ||
+        responseMessageTypeName.indexOf(namespacePackage) === 0
+      );
+    });
+  });
+}
+
 export function printFileDescriptorTSServices(fileDescriptor: FileDescriptorProto, exportMap: ExportMap) {
   if (fileDescriptor.getServiceList().length === 0) {
     return "";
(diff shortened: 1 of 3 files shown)


jonbretman:
@jonnyreeves added a test but not sure if there is a better way to test this than checking the generated source code for the import. Let me know what you think.

src/ts/fileDescriptorTSServices.ts
```diff
@@ -25,6 +41,10 @@ export function printFileDescriptorTSServices(fileDescriptor: FileDescriptorProt
 
   fileDescriptor.getDependencyList().forEach((dependency: string) => {
     const pseudoNamespace = filePathToPseudoNamespace(dependency);
+    // Don't output unused variables as that causes TS6133 error
+    if (!isUsed(fileDescriptor, pseudoNamespace, exportMap)) {
```
easyCZ: This can be moved into the functional iterator. We can write something like 
```
fileDescriptor.getDependencyList()
   .filter(dependency => isUsed(fileDescriptor, pseudoNamespace, exportMap))
   .forEach((dependency: string) => {...})
```

test/ts_test/src/service.ts
```diff
@@ -21,4 +23,11 @@ describe("ts service", () => {
     assert.strictEqual(simple_service_pb_service.SimpleService.DoStream.requestType, simple_service_pb.StreamRequest);
     assert.strictEqual(simple_service_pb_service.SimpleService.DoStream.responseType, external_child_message_pb.ExternalChildMessage);
   });
+  it("should not output imports for namespaces that are not used in the service definition", () => {
```
easyCZ: A better solution is to pull https://github.com/improbable-eng/ts-protoc-gen/blob/0135ec77326fb35535af0a7751d48c83eecea99e/src/ts/fileDescriptorTSServices.ts#L26-L34
into a separate method which accepts the `fileDescriptor`. We can then mock the `fileDescriptor` and assert the output string doesn't contain the import. 

I realize most of that testing style isn't in place so your solution will do for now and we can address it later.

src/ts/fileDescriptorTSServices.ts
```diff
@@ -5,6 +5,22 @@ import {FileDescriptorProto} from "google-protobuf/google/protobuf/descriptor_pb
 import {WellKnownTypesMap} from "../WellKnown";
 import {getFieldType, MESSAGE_TYPE} from "./FieldTypes";
 
+function isUsed(fileDescriptor: FileDescriptorProto, pseudoNamespace: string, exportMap: ExportMap) {
+  return fileDescriptor.getServiceList().some(service => {
+    return service.getMethodList().some(method => {
+      const requestMessageTypeName = getFieldType(MESSAGE_TYPE, method.getInputType().slice(1), "", exportMap);
+      const responseMessageTypeName = getFieldType(MESSAGE_TYPE, method.getOutputType().slice(1), "", exportMap);
+      if (requestMessageTypeName.indexOf(pseudoNamespace + ".") === 0) {
```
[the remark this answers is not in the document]
jonbretman: Updated.

src/ts/fileDescriptorTSServices.ts
jonbretman: Updated.

test/ts_test/src/service.ts
jonbretman: Agree this is a far better solution and happy to make a subsequent PR to try to improve the tests.

jonbretman:
@jonnyreeves @jonny-improbable @easyCZ Any plans to do a release now this has been merged?

jonny-improbable:
@jonbretman Yep, will aim pull together a 0.5.0 release this week

PR #40, opened 2018-03-01
Client interface that matches grpc-node

jonahbron:
This PR (based on @thakkarparth007's work) provides an interface in the generated service file that more closely matches that of the `grpc-node` implementation.  Instead of calling the grpc-web-client library directly, the developer can use their generated service like so:

```
import {GreeterClient} from "./helloworld_pb_service";
import {HelloRequest, HelloReply} from "./helloworld_pb";

const request = new HelloRequest();
request.setName('bobby');
const client = new GreeterClient('https://localhost:8443');
client.sayHello(request, undefined, (error, reply: HelloReply) => {
  console.log(reply.getMessage());
});
```

See the basic guide to GRPC-Node here for reference:

https://grpc.io/docs/tutorials/basic/node.html

Kudos to @thakkarparth007 for figuring out this solution and doing most of the work, this just generalizes his work and matches it to grpc-node.

While I'd love feedback on this PR, it's not quite ready to merge yet IMHO.  I want to add this behavior to the JS plugin (not just the TS plugin) and add unit tests for my changes.

PR #42, opened 2018-03-06
Added ability to generate typings for grpc-node service files

obto:
This PR adds the ability for users to output typings for the service files generated by protoc for node. Let me know if you have any questions or concerns about adding this to the library :)

Note: The addition of the two deps is strictly for the types they provide.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,12 +29,14 @@
     "google-protobuf": "^3.2.0"
   },
   "devDependencies": {
+    "@types/protobufjs": "^6.0.0",
     "@types/chai": "^3.4.35",
     "@types/google-protobuf": "^3.2.4",
     "@types/mocha": "^2.2.40",
     "@types/node": "^7.0.5",
     "babel": "^6.5.2",
     "chai": "^3.5.0",
+    "grpc": "^1.9.1",
     "mocha": "^3.2.0",
     "source-map-support": "^0.4.14",
     "tslint": "^4.5.1",
--- a/src/ts/fileDescriptorTSServices.ts
+++ b/src/ts/fileDescriptorTSServices.ts
@@ -76,3 +76,126 @@ export function printFileDescriptorTSServices(fileDescriptor: FileDescriptorProt
 
   return printer.getOutput();
 }
+
+export function printFileDescriptorTSGRPC(fileDescriptor: FileDescriptorProto, exportMap: ExportMap) {
+  if (fileDescriptor.getServiceList().length === 0) {
+    return "";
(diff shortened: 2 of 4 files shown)


anisimovsergey:
Exactly what I need!

jonny-improbable:
Thanks for contributing @obto; this looks great! Before I can merge this please could you follow up with the following changes:

1. Pull your new code out into a separate file, let's keep service generation separate
2. Add test coverage to ensure that the generated `.d.ts` are valid and don't regress with future changes (we have examples of this already for the grpc-web service definitions).

MOZGIII:
Just FYI, this seem to require the js files to be generated separately by the grpc-tools package.
I'm using @obto fork for now, looking forward to the merge.

Junkern:
What is the status of this PR?

In contrary to @jonahbron, I don't think it is rendered moot by #40. By using `service=true`I now (`0.6.0` version) indeed get types and stubs for my RPC calls, but the files have an `import {grpc} from "grpc-web-client"`. I am using the vanilla `grpc` dependency without any additional libraries (as it already has node.js support). 

The readme has a `Generating gRPC Service Stubs for use with grpc-web` entry but what about `Generating gRPC Service Stubs for use with vanilla-grpc``

Everything is working so far except that I cannot use the created `service.*` files due to not having the `grpc-web-client` dependency in my project. I could add a post-build step to remove the code I don't need, but I'd rather not go down that rabbit hole.

jonnyreeves:
@Junkern is correct, #40 generates service stubs for `grpc-web`, not google's own NodeJS grpc library.

This diff is going to require quite a lot of effort to get it rebased following a couple of extensive refactors (in particular #40 and #51) however I do see the benefit of having ts-protoc-gen output these typescript definitions.

I'm pretty busy at the moment and don't have a lot of time to contribute, is there anyone reading that would be in a position to pick this PR up, please?

obto:
Unfortunately, I wont have time to make the necessary changes (added tests and some refactoring) anytime soon. Quite swamped atm.

Junkern:
I could do it, as I want to have this anyway ;) Should I rework the current PR or start a new one (clean state....)?

jonnyreeves:
Awesome, thanks for volunteering.

Clean state referencing the work achieved here may be quicker IMHO.

Note you may wish to wait for #69 to land first.

Junkern:
FIY: #69 has landed, so I will start working on this

jonnyreeves:
Closing in favour of #75

PR #44, opened 2018-03-25
Create JavaScript sources and TypeScript definitions for grpc-web services

jonny-improbable:
This ensures that the final output is valid ES5 and doesn't require further transpiling; essential if you want your generated code to live in its own package and therefore not require further transpiling from TypeScript to JavaScript.

Removes all references to the `protoc-gen-js_service` as it is no longer required.


jonny-improbable:
This will also fix #32 (cc: @aberasarte)

PR #45, opened 2018-03-26
Stop using TypeScript Modules in grpc-web Service Definitions

### LINKED ISSUE #39: Consider deprecating the use of TypeScript namespaces to use ES2015 modules instead
The ability to leverage Babel 7 & TypeScript cannot happen with the current use of namespaces within the code `ts-protoc-gen` generates  for proto files.

The reason is that the use of `namespace` is problematic because it requires type information to transpile; a good article that goes over such a problem is [this one](http://artsy.github.io/blog/2017/11/27/Babel-7-and-TypeScript/)

Accordingly, there may be value to instead use ES6 modules for the use cases `namespace` is currently used by `ts-protoc-gen` to fix this.

### Steps to Reproduce
1. Setup Babel7 + Typescript (A quick starterkit that does this is [Microsoft's Typescript-Babel7 Starter Kit](https://github.com/Microsoft/TypeScript-Babel-Starter))
2. Set-up the necessary files to run `protoc` with `ts-protoc-gen` options
3. Attempt to compile the Typescript to JS
4. An error should show up saying something along the lines of the following snippet
```cmd
ERROR in .../foo/bar/baz_pb_service.ts
Module build failed: SyntaxError: /full/path/to/foo/bar/baz_pb_service.ts: Namespaces are not supported.
```

jonny-improbable:
Namespaces are not supported by Babel 7's TypeScript support; this PR changes the grpc-web Service TypeScript Definitions to no longer use them.

Fixes #39


src/ts/fileDescriptorTSServices.ts
```diff
@@ -144,18 +146,14 @@ export function printFileDescriptorJSServices(fileDescriptor: FileDescriptorProt
...
-        statics: {
-          methodName: `"${method.getName()}"`,
-          service: service.getName(),
-          requestStream: method.getClientStreaming(),
-          responseStream: method.getServerStreaming(),
-          requestType: requestMessageTypeName,
-          responseType: responseMessageTypeName,
-        }
-      });
+      printer.printLn(`${service.getName()}.${method.getName()} = {`);
```
jonny-improbable: Service Methods are now exported as an plain old javascript object instead of a 'class'

src/ts/fileDescriptorTSServices.ts
```diff
@@ -71,25 +71,27 @@ export function printFileDescriptorTSServices(fileDescriptor: FileDescriptorProt
...
-      methodPrinter.printLn(`export class ${method.getName()} {`);
-      methodPrinter.printIndentedLn(`static readonly methodName: string;`);
-      methodPrinter.printIndentedLn(`static readonly service: typeof ${service.getName()};`);
-      methodPrinter.printIndentedLn(`static readonly requestStream: ${method.getClientStreaming()};`);
-      methodPrinter.printIndentedLn(`static readonly responseStream: ${method.getServerStreaming()};`);
-      methodPrinter.printIndentedLn(`static readonly requestType: typeof ${requestMessageTypeName};`);
-      methodPrinter.printIndentedLn(`static readonly responseType: typeof ${responseMessageTypeName};`);
-      methodPrinter.printLn(`}`);
+
+      printer.printLn(`type ${service.getName()}${method.getName()} = {`);
```
jonny-improbable: We now export a `type` for each Service Method in the TypeScript Definition, these types are used to provide the type hint for each exported method on the Service Class (printed below)

PR #53, opened 2018-04-02
Remove invalid npm script commands

easyCZ:
Removes `npm lint` commands from inside the `test` directory. These scripts cannot be run as they have a dependency on `tslint` which is a dependency of the parent and therefore unavailable unless installed in global scope.


easyCZ:
This is now redundant in favour of #55

PR #54, opened 2018-04-02
Add CONTRIBUTING.md

easyCZ:
Add a CONTRIBUTING.md file with description of the steps to take to ensure the changes are compatible. This is a first pass and the guidelines will need to be updated in the future.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,15 @@
+# Contributing
+
+Firstly, thanks for taking the time to contribute.
+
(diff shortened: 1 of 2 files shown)


jonny-improbable:
Nice, thanks for adding this!  Can I suggest we also bring in a Code of Conduct section as well? We can just grab the one from [this example](https://gist.github.com/PurpleBooth/b24679402957c63ec426) to begin with.

easyCZ:
Added.

PR #55, opened 2018-04-02
Move generated protos to examples and check them in

easyCZ:
In most use cases of `ts-protoc-gen`, the generated files will be part of the repository. I believe we should also follow this suite and check in our test protos.

## Why?
* Visibility. A potential user of `ts-protoc-gen` wants to be able to see what the generated files look like before having to try it out. We should make this possible.
* Examples - by having the protos checked in, we can write both examples for given compilation targets (ie. node, grpc-web). This makes it easier to onboard new users.
* Testing - with the above examples, we have automated tests against grpc-web and grpc-node to ensure we are not making breaking changes. We can also experiment with usage patterns in those frameworks and we can enable the community to help us build up examples.
* Multiple frameworks support - This makes the transition into having multiple output targets generated out of `ts-protoc-gen` in one place and have sufficient testing scaffolding

## Consequences
* Maintainability becomes more difficult. We need to ensure the generated protos match the original ones and make corresponding changes. This can of course be automated and become part of CI.

## Changes in this PR
* Move our test `proto` directory to the root of the repo
* Update the `generate.sh` script to generate into a `examples` directory
* Flatten our test directory by removing nesting
* Flatten nested `package.json` definitions and consolidate them inside root `package.json`
* Use `ts-node` when running tests to avoid having to compile TS each time we're running tests
* Fix a few lint issues


jonny-improbable:
@easyCZ could you bring this up to speed with the latest master please?

easyCZ:
Rebased & fixed issues. Let me know what you think

generate.sh
```diff
@@ -0,0 +1,29 @@
...
+
+echo "Ensuring we have NPM packages installed..."
+npm install
+
+echo "Compiling ts-protoc-gen..."
+npm run build
+
+echo "Generating proto definitions..."
+
+if [ -d $EXAMPLES_GENERATED_DIR ]
```
jonnyreeves: nit: Wrap `$EXAMPLES_GENERATED_DIR` in quotes to avoid potential issues with spaces, ie:

```
+if [ -d "$EXAMPLES_GENERATED_DIR" ]

```

jonnyreeves:
Hey @easyCZ, I've just taken another look over this and it LGTM; sorry to make you rebase again (the effort to fix the tests in #63 and #64 have caused some conflicts) but once complete let's land this.

As a future improvement I would suggest that we generate have a test on CI which checks that the generated code matches what's checked in to avoid them going out of sync - there is something similar in grpc-web for the generated documentation [here](https://github.com/improbable-eng/grpc-web/blob/master/go/fixup.sh#L10)

easyCZ:
@jonnyreeves Thanks, this is now rebased. The CI check is definitely something to add but I'd defer that until a separate PR

PR #56, opened 2018-04-02
Fix replace all bug in filePathFromProtoWithoutExtension

easyCZ:
This PR fixes a bug which occurs for users storing their `proto` definitions inside a `.proto` directory.

The bug is caused by `filePathFromProtoWithoutExtension` replacing all occurrences of `.proto` when renaming a file instead of just the file suffix.

A drive-by improvement to the documentation of ts-protoc-gen entry point.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,16 +1,19 @@
-/**
- * This is the ProtoC compiler plugin.
- *
- * It only accepts stdin/stdout output according to the protocol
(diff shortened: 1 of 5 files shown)


jonnyreeves:
LGTM, although I propose we move unit tests to sit alongside source files with a .spec.ts extension.

This would require a new npm script (npm run test:unit) and a change to the .npmignore to exclude *.spec.ts from being published to npm.

LMKWYT.

jonny-improbable:
@easyCZ did you want me to land this as-is - happy to do so.

PR #60, opened 2018-05-03
Adds full response object as third argument to rpc callback

jonbretman:


jonnyreeves:
Thanks for raising this patch.

In a similar vein to the discussion in #59, I'm not currently comfortable with us breaking away from the pattern/API present in the official node-grpc library.

Note that the official node-grpc API returns a client object from unary invocations which will dispatch an event when the metadata becomes available, perhaps you would consider implementing this in which was I would gladly land it.

Thanks

PR #61, opened 2018-05-09
Don't use reserved words as function names in generated service

jonbretman:
* Function names in the generated service code are run through `normaliseFieldObjectName` to avoid invalid syntax
* Test added to check this behaviour
* Make sure all tests run and fix some import path errors in `test/service/grpcweb.ts`


jonnyreeves:
Thanks for raising 😀

`delete` is but one of many reserved keywords in JavaScript. We already have a utility which normalises field names in utils.ts (`normaliseFieldObjectName`) - please could you modify your patch to handle all of these keywords in a generic fashion and add some test coverage to ensure we don't regress on this change in the future?

Thanks

jonbretman:
@jonnyreeves so it looks like https://github.com/improbable-eng/ts-protoc-gen/pull/56 actually stopped the tests from running 😢You can see in the Travis build that 0 test ran - https://travis-ci.org/improbable-eng/ts-protoc-gen/builds/361377789?utm_source=github_status&utm_medium=notification

```
> mocha --require source-map-support/register ./build/src
  0 passing (3ms)
```

My guess is it's [this line](https://github.com/improbable-eng/ts-protoc-gen/pull/56/files#diff-bee6534dd0177f56c1d2dde31c661a07R2).

I've add a look at trying to fix them but getting a failing test - https://github.com/improbable-eng/ts-protoc-gen/pull/62

jonbretman:
@jonnyreeves ok all fixed now - using the `normaliseFieldObjectName` util function. Also only some of the tests were running so fixed that.

test/proto/examplecom/simple_service.proto
```diff
@@ -12,11 +12,16 @@ message UnaryRequest {
...
 message StreamRequest {
   string some_string = 1;
 }
 
 service SimpleService {
   rpc DoUnary(UnaryRequest) returns (othercom.ExternalChildMessage) {}
   rpc DoStream(StreamRequest) returns (stream othercom.ExternalChildMessage) {}
+
+  // checks that rpc methods that use reserved JS words don't generate invalid code
+  rpc Delete(UnaryRequest) returns (UnaryResponse) {}
```
jonnyreeves: Could you please add a test to ensure methods are appropriately prefixed?

jonnyreeves:
Morning @jonahbron, the example protos have been moved by the change landed in #55, you will need to rebase your changes. Also, as requested, please can you add a test-case so we don't accidentally regress on this feature in the future?

Thanks :)

jonnyreeves:
Thanks @jonbretman; I'll look to roll out a new release of ts-protoc-gen at some point this week; hopefully we can also get the fix for #67 in as well.

Issue #63, opened 2018-05-09
Camel-cased OneOf message typings are not generated correctly.


MHDante:
There's an issue with the TS generated for OneOfs that are camelcased. Both the Case Enum type and the Case enum accessor have the wrong casing. 

The following message:

```protobuf
message MyMessage{
    oneof thisIsAField{
        int32 anInt = 1;
        string theString = 2;
    }
}
```

This produces the following TS code:

```typescript
export namespace MyMessage {
  export type AsObject = {
    anint: number,
    thestring: string,
  }

  export enum ThisIsAFieldCase {
    THISISAFIELD_NOT_SET = 0,
    ANINT = 1,
    THESTRING = 2,
  }
}
```

And the following JS code:
```javascript
/**
 * @enum {number}
 */
proto.MyMessage.ThisisafieldCase = {
  THISISAFIELD_NOT_SET: 0,
  ANINT: 1,
  THESTRING: 2
};
```

As far as I can tell, there is a general bug with the way camelcase fields are handled in pbjs, but it sees like ts was always using the same bug (and was therefore compatible). This is not the case with OneOfs

As a workaround, you can either cast to Any, or you can use one word oneof fields.

jonnyreeves:
Thanks for raising @MHDante, Please could you take a look at #67 and confirm this addresses the main issue here (the classname not matching).

If I understand correctly there is a secondary issue that the fieldNames in the `AsObject` representation also do no match.

MHDante:
As far as I can tell, there is no secondary issue in the AsObject naming. (Other than protoc's inherent destruction of camel humps) 

#67 does in fact address this issue here. Thanks for your prompt attention.

MHDante:
Out of curiosity, is there an issue regarding the loss of camelcasing in the JS compiler? The closest thing I found was: https://github.com/google/protobuf/issues/1591

PR #67, opened 2018-05-20
Bugfix/63 oneof casing

jonnyreeves:
Repro case for #63


jonnyreeves:
Sorry for the delay here, I encountered some build issues when I pulled the latest master which I've looked to address in #69. Once that is landed I will bring this up to speed and look to get it merged.

PR #69, opened 2018-05-22
Fix and Restructure Tests

jonnyreeves:
Things got a little bit out of sync after #55 landed on master and we lost some of the safe-guards that were introduced in #64.

This PR refactors the layout of the tests, making a split between integration and unit tests with an npm script for each. It also re-instates the 'check-mocha-report' check to ensure that at least 1 test passed.

I've committed a small cardinal sin by doing some drive-by linting fixes that were bugging me, sorry it makes the diff a little chunkier than it needed to be.


package.json
```diff
@@ -8,7 +8,9 @@
     "clean": "rm -rf lib",
     "build": "npm run clean && tsc",
     "generate": "./generate.sh",
-    "test": "mocha --recursive -r ts-node/register -r source-map-support/register 'test/**/*.ts'",
+    "test": "npm run test:unit && npm run test:integration",
```
jonnyreeves: @easyCZ this is the bulk of the change which consists of 2 parts:

1. Change the layout of the tests on disk making a clear split between integration and unit tests. This allowed me to remove the `test/test_ts` folder which contained a single unit test, some helper code (used by the integration tests) and some scripts for asserting the output from mocha (see point 2 below).

2. Invoke the new `mocha-run-suite.sh` script which invokes mocha and then performs a quick assertion to check that at least 1 test passed. Without this check it's possible for zero tests to run but mocha to still return exit code 0 marking the build as green when it shouldn't be.

PR #80, opened 2018-05-28
Download protoc when generating protos

### LINKED ISSUE #78: Inconsistent setField method in generated examples
A question about the `examples/generated` files. 

in #67, @jonnyreeves commited updated the generated files when commiting, resulting in the following change:

https://github.com/improbable-eng/ts-protoc-gen/pull/67/files#diff-6cee6eac2b0bae17197dc838b7af57a2L208

I'm trying to set up development on my local environment, and the most recent version of protoc (3.5.1) is generating js that uses the  `setProto3*Field` methods (e.g. `setProto3EnumField`)

I asume that perhaps @jonnyreeves'  protoc is out of date, or is there a way to generate the js file using the generic `setField` method?

easyCZ:
* Add a download step to fetch protoc from github
* Use it for generating protos
* Clean up when generate script exits
* Regenerates examples

Fixes #78

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,6 @@ before_install:
 - sudo dpkg -i bazel_0.11.0-linux-x86_64.deb
 install:
 - sudo apt-get install unzip
-- bash install-protobuf.sh
 - export PATH=$TRAVIS_BUILD_DIR/protobuf/bin:$PATH
 - npm install
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: node_js
-env:
-- PROTOBUF_VER="3.2.0"
 before_deploy: |
   PKG_VERSION=$(node -p "require('./package.json').version")
   SHORT_COMMIT_HASH=$(git rev-parse --short HEAD)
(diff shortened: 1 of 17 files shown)


generate.sh
```diff
@@ -5,12 +5,34 @@ set -e
...
+esac
+echo "You appear to be running on ${platform}"
+
 echo "Ensuring we have NPM packages installed..."
 npm install
 
 echo "Compiling ts-protoc-gen..."
 npm run build
 
+echo "Downloading protoc v3.5.1... for ${platform}"
```
jonnyreeves: Suggest we read the version from a variable to make it easier to upgrade

jonnyreeves:
Thanks! Nice fix

generate.sh
easyCZ: Good shout! Fixed.

easyCZ:
@jonnyreeves Fixed the comments, thanks for pointing out the usage in travis. I've actually removed the `install-protobuf.sh` script as it is now redundant

jonnyreeves:
Nice, please could you:

1. Remove `PROTOBUF_VER="3.2.0"` from the `travis.yaml` as it's no longer required.
2. consider adding `/protoc` to `.gitignore` just incase?

.travis.yml
```diff
@@ -37,7 +37,6 @@ before_install:
 - sudo dpkg -i bazel_0.11.0-linux-x86_64.deb
 install:
 - sudo apt-get install unzip
-- bash install-protobuf.sh
 - export PATH=$TRAVIS_BUILD_DIR/protobuf/bin:$PATH
```
jonnyreeves: @easyCZ I believe this `PATH` statement was only used to locate the `protoc` bin unzipped during `install-protofbuf.sh` and can safely be removed

easyCZ: 👍 Also removed install of unzip

jonnyreeves:
🛬

PR #82, opened 2018-05-29
Implement Client Streaming and BiDi Streaming for grpc-web

### LINKED ISSUE #66: Support client-side and bidi streaming in grpc-web Service Stubs
I'm trying to use this library on node.js to talk to a gRPC server. In my proto file I have 
```
rpc WatchMetric(stream MetricSubscriptionRequest) returns (stream Event) {}
```

However, this results in the following code being generated:

*example_pb_service.js*
```
Watcher.prototype.watchMetric = function watchMetric() {
  throw new Error("Client streaming is not currently supported");
}
```

*example_pb_service.d.ts*
```
watchMetric(): void;
```

Since v0.6.0, grpc-web provides an experimental websocket transport which should allow client-side and bidi streaming.

jonnyreeves:
Fixes #66 

This PR provides support for both Client Streaming and BiDirectional Streaming to the generated service stubs.

As an additional drive-by, I've also provided support for debugging the tests using a connected Chrome Inspector/Debugger.


jonny-improbable:
@pjm17971 if could have some time to review that would be awesome, thanks!

test/helpers/fakeGrpcTransport.ts
```diff
@@ -11,6 +11,14 @@ function frameResponse(request: Message): Uint8Array {
   return new Uint8Array(frame);
 }
 
+export function frameRequest(request: Message): ArrayBufferView {
```
easyCZ: [nit] Could you please add a comment on why this is needed? This will allow external contributors to get up to speed with the code base easier.

jonnyreeves: Oops sorry, missed this, will backfill :) <3

jonnyreeves:
@pjm17971 If you can, could you please try upgrading to 0.7.7-pre.5959599 of ts-protoc-gen and give some feedback on how you get on?

PR #84, opened 2018-05-29
Add Custom Bazel rule

coltonmorris:
This will add a custom bazel rule, which uses the `protoc-gen-ts` binary for generation!

Some notes:
1. Changed import paths of protos within the `proto` directory to better align with bazel.
    - This change makes the generated code output into another subdirectory named `proto`. This might seems a bit ugly, but I believe this is better for long term sanity.
2. gg


coltonmorris:
bitmoji
WOMM, I'll figure that out

jonnyreeves:
Can't say I'm fluent in bazel but I'm going to assume this is a positive change for the wider community and will land it. 🤘🚀

PR #92, opened 2018-06-23
Add compatibility with rules_typescript

Globegitter:
This was taken from: https://github.com/bazelbuild/rules_typescript/blob/d25f3ce7438d8fb3497115f04a029232c8ef2861/internal/protobufjs/ts_proto_library.bzl#L93 and adapted for this rule

Edit: Just to clarify this was necessary for me to get grpc-web work with rules_typescript.

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -50,7 +50,7 @@ def _typescript_proto_library_impl(ctx):
     proto_inputs.append(normalized_file)
     append_to_outputs(ctx, file_name, js_outputs, dts_outputs, file_modifications)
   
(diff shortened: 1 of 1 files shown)


alexeagle:
Ideally we should generate different JS content for devmode "es5" vs prodmode "es6" so that bundlers can do the best job on them, like we do in `ts_proto_library`

jonnyreeves:
@alexeagle what is the status of this change? Is it still in progress or should it be landed?

Globegitter:
Seems like this can be closed in favour of #138

PR #104, opened 2018-08-31
Add support for jstype annotation

### LINKED ISSUE #17: Translating int64 to TypeScript Number type sacrifices accuracy
The [TypeScript Number type](https://www.typescriptlang.org/docs/handbook/basic-types.html) is a floating point number, which means attempting to express the int64 range in numbers will lead to inaccuracies. The protoc compiler therefore uses strings when expressing large numeric types in JSON.

Perhaps something similar should be considered for the TypeScript type translation?

jonny-improbable:
Required an upgrade to google-protobuf 3.6.1, otherwise very straight forward.

Fixes #17


jonny-improbable:
cc @cmoad and @johanbrandhorst - let me know if that works for you.

src/ts/message.ts
```diff
@@ -104,7 +108,20 @@ export function printMessage(fileName: string, exportMap: ExportMap, messageDesc
         exportType = filePathToPseudoNamespace(fieldEnumType.fileName) + "." + withinNamespace;
       }
     } else {
-      exportType = getTypeName(type);
+      if (field.getOptions() && field.getOptions().hasJstype()) {
```
jonny-improbable: Sorry for making this code even more disgusting, I do have a local branch which contains a large refactor to message parsing - I'd say it's about 75% complete .

johanbrandhorst:
Which types is this supported on?

cmoad:
@johanbrandhorst The 5 types that end in "64".

- FieldDescriptorProto.Type.TYPE_INT64
- FieldDescriptorProto.Type.TYPE_UINT64
- FieldDescriptorProto.Type.TYPE_SINT64
- FieldDescriptorProto.Type.TYPE_FIXED64
- FieldDescriptorProto.Type.TYPE_SFIXED64

johanbrandhorst:
Nice!

johanbrandhorst:
Not sure we have reason to celebrate just yet unfortunately: https://github.com/protocolbuffers/protobuf/blob/master/js/binary/decoder.js#L724 :(. Seems like they've just not bothered with some types still.

johanbrandhorst:
Might want to look through the rest of that file before deciding it doesn't apply to you, I can't remember but there might be more integer types that are unsupported. Might be worth writing some tests for? Anyway, I agree of course that merging this is a step in the right direction.

jonny-improbable:
Thanks for taking a look @cmoad and @johanbrandhorst.

@johanbrandhorst I'm confused by your comment around ZigZag encoding as my tests are passing 😕 Quite possible that I've just misunderstood tho :)

src/ts/message.ts
jonny-improbable: Done, thanks for the suggestion.

johanbrandhorst:
I think a more appropriate test would be to write `MaxInt64` to a variable, send it to a Go server, verify that it's == `math.MaxInt64`, send it back to the client and again verify that it's == `MaxInt64`. I'm not sure what your test really tests at the moment.

For reference, I manually implemented this in my GopherJS gRPC-Web library when I realised it couldn't be done with the JS bindings:

https://github.com/johanbrandhorst/protobuf/blob/227432c61df7f8b5bace4bb8064797c2aea8dc28/jspb/reader.go#L64

https://github.com/johanbrandhorst/protobuf/blob/227432c61df7f8b5bace4bb8064797c2aea8dc28/jspb/writer.go#L62

With tests:

https://github.com/johanbrandhorst/protobuf/blob/227432c61df7f8b5bace4bb8064797c2aea8dc28/test/client/main.go#L122

ZigZagInt == SINT64. https://developers.google.com/protocol-buffers/docs/encoding#signed-integers.

johanbrandhorst:
I reiterate that I welcome this change, I just want everyone involved to be aware that this doesn't fix the 64bit issue, unfortunately, through no fault of this library. Let's get this in and hope they will sort the protobuf library soon :)

jonny-improbable:
Thanks both

PR #108, opened 2018-09-10
Be explicit about the need for a CommonJS environment.

jonnyreeves:


jonny-improbable:
@easyCZ another cheeky lil PR ❤️

PR #124, opened 2018-10-15
Allow canceling unary calls

virtuald:
- Fixes https://github.com/improbable-eng/grpc-web/issues/252

PR #125, opened 2018-10-15
Transport member of ServiceClientOptions should be optional

virtuald:
What the title says. Should be good, given that clients are initialized with...

```
function BookServiceClient(serviceHost, options) {
  this.serviceHost = serviceHost;
  this.options = options || {};
}
```


jonnyreeves:
Thanks for contributing @virtuald! I'm happy to approve this change, but you also need to re-generate the example code. You can achieve this by invoking `./generate.sh`.

The fact that CI passes implies that we are missing an automated check to ensure that the example code is in sync; I'll add a task to track this improvement.

jonny-improbable:
Same as #124, @virtuald could you please rebase your changes against the latest master, this should highlight the discrepancy with the generated code being out of sync.

PR #127, opened 2018-10-16
CI should check that generated code has been committed

jonny-improbable:
Add a travis build step which re-generates the generated code and then checks for changes to the workspce. If any are detected it will fail the build and prompt the user to run the `./generate.sh` script.

### DIFF
--- a/travis-ci-build.sh
+++ b/travis-ci-build.sh
@@ -3,4 +3,12 @@ set -ex
 
 npm run lint
 npm test
+
+./generate.sh
+MODIFIED_FILES=$(git diff --name-only)
+if [[ -n $MODIFIED_FILES ]]; then
+  echo "ERROR: Changes detected in generated code, please run './generate.sh' and check-in the changes."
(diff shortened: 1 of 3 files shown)


jonnyreeves:
Pretty sure we are seeing these errors because it's the first time this script has been executed on Travis.

I'll try switching to using unzip instead of tar

travis-ci-build.sh
```diff
@@ -3,4 +3,12 @@ set -ex
 
 npm run lint
 npm test
+
+./generate.sh
+MODIFIED_FILES=$(git diff --name-only)
+if [[ -n $MODIFIED_FILES ]]; then
+  echo "ERROR: Changes detected in generated code, please run './generate.sh' and check-in the changes."
```
easyCZ: This will unfortunately be invisible to our users and hence they won't be able to self-fix this. Not sure if we have a fix for this. We could make the project visible but not modifiable to the public?

PR #129, opened 2018-10-17
Update bazel to use fine grained dependencies and cleaned up installation

coltonmorris:
Using fine grained dependencies now! This also allowed me to clean up the installation process to no longer need to load in `typescript_proto_dependencies`. 

`typescript_proto_library` now depends on nodejs_binary target rather than the npm binary.


package.json
```diff
@@ -32,6 +32,7 @@
     "google-protobuf": "^3.6.1"
   },
   "devDependencies": {
+    "@bazel/typescript": "^0.20.3",
```
jonnyreeves: you'll need to perform an `npm install` (using npm v6) to propagate this change through to the `package-lock.json` file.

coltonmorris: I might be looking at github ui wrongly, but I believe I did that already: 
https://github.com/improbable-eng/ts-protoc-gen/blob/9e831dedc3a359882f8ff4f84618a85dd4eba914/package-lock.json#L7

jonnyreeves: Oh yes, so you have, sorry about that.

The build is failing because the `package-lock.json` is modified after the `npm install` invocation inside `generate.sh`.

Perhaps the version of npm being used differs between your local setup and on Travis?

coltonmorris: 👍

PR #131, opened 2018-10-20
Correct generated client, bidi stream method stubs

### LINKED ISSUE #132: Client streaming stub generation uses wrong type
It seems to be using the responseType instead of the requesType, see https://github.com/improbable-eng/ts-protoc-gen/blob/51f57eaaafa0434433c218e40a7649fe095a46ea/src/service/grpcweb.ts#L534

johanbrandhorst:
It appears we were using the responseType instead of the requestType when creating the stream.

Fixes #132, #133

### DIFF
--- a/src/service/grpcweb.ts
+++ b/src/service/grpcweb.ts
@@ -222,13 +222,13 @@ function generateTypescriptDefinition(fileDescriptor: FileDescriptorProto, expor
   printer.printIndentedLn(`on(type: 'end', handler: () => void): RequestStream<T>;`);
   printer.printIndentedLn(`on(type: 'status', handler: (status: Status) => void): RequestStream<T>;`);
   printer.printLn(`}`);
-  printer.printLn(`interface BidirectionalStream<T> {`);
-  printer.printIndentedLn(`write(message: T): BidirectionalStream<T>;`);
+  printer.printLn(`interface BidirectionalStream<ReqT, ResT> {`);
+  printer.printIndentedLn(`write(message: ReqT): BidirectionalStream<ReqT, ResT>;`);
   printer.printIndentedLn(`end(): void;`);
   printer.printIndentedLn(`cancel(): void;`);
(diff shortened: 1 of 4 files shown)


johanbrandhorst:
@jonny-improbable should be a simple merge I hope.

johanbrandhorst:
I've found another problem, lets wait with merging this please.

johanbrandhorst:
@jonny-improbable ok there we go.

src/service/grpcweb.ts
```diff
@@ -222,13 +222,13 @@ function generateTypescriptDefinition(fileDescriptor: FileDescriptorProto, expor
   printer.printIndentedLn(`on(type: 'end', handler: () => void): RequestStream<T>;`);
   printer.printIndentedLn(`on(type: 'status', handler: (status: Status) => void): RequestStream<T>;`);
   printer.printLn(`}`);
-  printer.printLn(`interface BidirectionalStream<T> {`);
-  printer.printIndentedLn(`write(message: T): BidirectionalStream<T>;`);
+  printer.printLn(`interface BidirectionalStream<Req, Res> {`);
```
johanbrandhorst: I'd like to know opinions on the Generic type variables here, but I can't use `<R, R>`, and I think `<T, V>` is just unclear.

johanbrandhorst: Maybe `<ReqT, ResT>`?

johanbrandhorst:
My formatter has formatted the files I've touched, I can try and revert those changes and keep the edits minimal if required.

src/service/grpcweb.ts
jonnyreeves: Yep ReqT and ResT work for me

johanbrandhorst:
@jonnyreeves should be good now

jonnyreeves:
Both work 😁

jonnyreeves:
Published as (0.7.8-pre.217169b)

PR #146, opened 2019-01-09
generate clients that use promises for unary methods

vilterp:
A simple way of generating clients that can be used in a more idiomatic (promise-based) way. Missing things like support for methods other than unary, and passing metadata.

This may not be ready for merging into master, but supports my needs at the moment, so I may be using this fork for a while.

### DIFF
--- a/examples/generated/proto/examplecom/simple_service_pb_service.d.ts
+++ b/examples/generated/proto/examplecom/simple_service_pb_service.d.ts
@@ -115,3 +115,15 @@ export class SimpleServiceClient {
   ): UnaryResponse;
 }
 
+export class SimpleServicePromisesClient {
+  readonly serviceHost: string;
+
+  constructor(serviceHost: string, options?: grpc.RpcOptions);
+  doUnary(
+    requestMessage: proto_examplecom_simple_service_pb.UnaryRequest,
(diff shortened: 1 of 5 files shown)


jonnyreeves:
Hi @vilterp, thanks for raising awareness of your fork.

I've added a comment to #59 which provides some closure to the support for generating clients that use Promises and Observables:

> As discussed in a handful of other issues, it's my current belief that ts-protoc-gen should be built around callbacks and event handlers. Other abstractions on top of these concepts (or: Promises and Observables) can be created by the community, presumably as a wrapper, ie:
> 
> ```
> promisify(myClient.doUnary, myMetadata)
>   .then(response => /* … */)
>   .catch(err => /* … */)
> ```
> My rationale is that I would like to keep ts-protoc-gen's API surface area as small as possible to make it easier to maintain.

Once again thank you for raising awareness and should you agree and be willing to create a new standalone package I would be more than happy to reference it in the ts-protoc-gen README.

vilterp:
Thanks for your response @jonnyreeves; I understand the desire to keep the API small and compliant with node-gRPC. Will consider making the `promisify` wrapper or a standalone package.

apeace:
@vilterp we forked off the `v0.10.0` release and added your commit on top: https://github.com/PilotFiber/ts-protoc-gen

We're planning to rebase from time to time whenever there is a new release.

Issue #156, opened 2019-03-29
Handling of enums


mattvagni:
I've noticed something quite strange about the handling of enums, I'm not sure what the solution is but am happy to help contribute a fix once decided what the correct behaviour should be (and whether this is even an issue).

Assuming an enum in proto file like this:

```proto
enum Emotion {
  POSITIVE = 0;
  NEUTRAL = 1;   
  NEGATIVE = 2; 
}
```

The typescript generation will output:

```tsx
enum Emotion {
  POSITIVE = 0,
  NEUTRAL = 1,   
  NEGATIVE = 2, 
}
```

The matching javascript output is:

```js
/**
 * @enum {number}
 */
proto.stock.Emotion = {
  POSITIVE: 0,
  NEUTRAL: 1,
  NEGATIVE: 2,
};
```

### The problem

The problem is that typescript enums should actually be outputted like so in javascript: 

```js
var Emotion;
(function (Emotion) {
    Emotion[Emotion["POSITIVE"] = 0] = "POSITIVE";
    Emotion[Emotion["NEUTRAL"] = 1] = "NEUTRAL";
    Emotion[Emotion["NEGATIVE"] = 2] = "NEGATIVE";
})(Emotion || (Emotion = {}));
```

This resulting value of the `Emotion` var becomes this once the above code is evaluated:

```js
var Emotion = {
  0: "POSITIVE"
  1: "NEUTRAL"
  2: "NEGATIVE"
  NEGATIVE: 2
  NEUTRAL: 1
  POSITIVE: 0
}
```

This allows for mapping in both directions so that [both of these would be truthy](https://www.typescriptlang.org/play/#src=enum%20Emotion%20%7B%0D%0A%20%20POSITIVE%20%3D%200%2C%0D%0A%20%20NEUTRAL%20%3D%201%2C%20%20%20%0D%0A%20%20NEGATIVE%20%3D%202%2C%20%0D%0A%7D%0D%0A%0D%0Aconsole.log(Emotion%5B0%5D%20%3D%3D%3D%20'POSITIVE')%3B%0D%0Aconsole.log(Emotion.POSITIVE%20%3D%3D%3D%200)%3B):

```js
console.log(Emotion[0] === 'POSITIVE');
console.log(Emotion.POSITIVE === 0);
```

If we treat the javascript output from protoc as correct it means that potentially this plugins' handling of proto enums should be different.

### Potential solutions

**1. Output type definitions that match the outputted javascript**

[...]

PR #157, opened 2019-03-30
Handling of enums

mattvagni:
More detail in: https://github.com/improbable-eng/ts-protoc-gen/issues/156

## Changes

- This changes how the type generation handles protobuf enums
- Regenerated all examples to run tests against

## Verification

I checked and there seem to be a fair few tests for enum handling, the code changes are backwards compatible so they were not modified. Not sure if there are any additional tests worth writing? I did however have to change some of the casting in the integration test for maps - not sure if the changes makes sense though 🤔

### DIFF
--- a/examples/generated/proto/examplecom/enum_message_pb.d.ts
+++ b/examples/generated/proto/examplecom/enum_message_pb.d.ts
@@ -5,21 +5,21 @@ import * as jspb from "google-protobuf";
 import * as proto_othercom_external_enum_pb from "../../proto/othercom/external_enum_pb";
 
 export class EnumMessage extends jspb.Message {
-  getInternalEnum(): EnumMessage.InternalEnum;
-  setInternalEnum(value: EnumMessage.InternalEnum): void;
+  getInternalEnum(): EnumMessage.InternalEnumMap[keyof EnumMessage.InternalEnumMap];
+  setInternalEnum(value: EnumMessage.InternalEnumMap[keyof EnumMessage.InternalEnumMap]): void;
 
   clearInternalEnumsList(): void;
-  getInternalEnumsList(): Array<EnumMessage.InternalEnum>;
-  setInternalEnumsList(value: Array<EnumMessage.InternalEnum>): void;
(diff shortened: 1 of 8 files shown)


mattvagni:
Haha brave 😅 Ok, thanks for merging this so quickly.

jonny-improbable:
Fortune favours the brave... plus we can rely on feedback from the community if anything goes proper bang :D

You can dogfood it yourself by upgrading to `0.9.1-pre.901b08e` via npm.

mahileeb:
Hi,

this change is causing compile errors for our app where we use the enum as a field type, e.g. the mode field in the below

```
type OwnProps = {
    name?: string;
    mode?: EnvironmentMode;
    version?: string;
    build?: string;
    className?: string;
}
```
causes

```(8,12): 'EnvironmentMode' refers to a value, but is being used as a type here.```

If I use the generate EnvironmentModeMap interface then it will compile

mattvagni:
@mahileeb Could it be that you are importing `EnvironmentMap` object from the generated js client, not the typescript type? 

Otherwise an example repo/gist would be helpful in order for me to help.

mahileeb:
Hmm, the import looks like:

```import { EnvironmentMode } from "protobuf/generated/com/mahifx/model/echo_protomap_pb";```

Is that the correct way to import it? Here's gist of the proto schema and the usage: https://gist.github.com/mahileeb/1ce0c625634d10595351c1110711901f

jacobfederer:
Any update on this issue? This is the dealbreaker for upgrading to 0.10 and higher.

yoyomolinas:
With this PR #157  many of us are unable to use Protobuf enums with Typescript. As pointed out in #210, https://github.com/improbable-eng/ts-protoc-gen/issues/187 and https://github.com/improbable-eng/ts-protoc-gen/issues/268, I think this issue needs further attention.  Any clarification appreciated.

thesayyn:
There is an alternative plugin that you can use to solve this. 
See: https://github.com/thesayyn/protoc-gen-ts/issues/39

screenpunch:
We're also stuck with 0.9, this change has made protobuf enums unusable for us also.

PR #158, opened 2019-03-31
Add option to general Service Clients or just Stubs.

jonny-improbable:
## Changes
Allow consumers to choose if they wish to generate full Service Clients (for use with grpc-web) or just the minimal boilerplate (stubs) required when invoking grpc-web directly.

The primary driver here is to reduce overall bundle-size. Tree shaking is not an option as the generated code use CommonJS modules, not ES6 ones.

## Verification
- Manual verification.


src/service/grpcweb.ts
```diff
@@ -201,47 +202,21 @@ function generateTypescriptDefinition(fileDescriptor: FileDescriptorProto, expor
     });
 
 
+  if (options.serviceType === "client") {
+    printClientInterfaces(printer);
 
-  printer.printLn(`export type ServiceError = { message: string, code: number; metadata: grpc.Metadata }`);
```
jonny-improbable: Pulled this block out into it's own method.

Issue #155, opened 2019-03-21
Server Streams `onStatus` happens after `onEnd`, would expect the reverse


hectim:
**Versions of relevant software used**
0.9.0

**What happened**
I am writing a promise wrapper around the generated grpc client requests. I thought it made sense to resolve the promise inside the `onEnd` handler, but found that if my call failed it would hit the `onEnd`, resolve the promise, and then get a status code other than 0 (meaning my call actually failed).

**What you expected to happen**
I would expect the `end` handler to be called very last. Additionally, I believe it would be nice to pass the status to the `end` handler. Since the current implementation doesn't take any parameters, adding an optional status parameter should be a pizza cake. 

**How to reproduce it (as minimally and precisely as possible)**:
```typescript
const streamRequest = myClient.myRequest();
streamRequest.on('status', () => console.log('happens second'));
streamRequest.on('end', () => console.log('happens first'));
await streamRequest;
```

***

Am I missing something major here or was it just a small implementation detail? Would you be open to a PR that essentially just swaps the order the listeners are called in the onEnd?
```typescript
onEnd: function (status, statusMessage, trailers) {
  listeners.end.forEach(function (handler) {
    handler();
  });
  listeners.status.forEach(function (handler) {
    handler({ code: status, details: statusMessage, metadata: trailers });
  });
  listeners = null;
}
```
and updates the end types for an optional parameter from this:
``` typescript
on(type: 'end', handler: () => void): ResponseStream<T>;
```
to this:
```typescript 
on(type: 'end', handler: (status?: Status) => void): ResponseStream<T>;
```

jonnyreeves:
Sorry for the slow response @hectim. Yes to both of these things please :)

PR #159, opened 2019-04-02
Swap onStatus to invoke BEFORE onEnd rather than after

hectim:
## Changes

implemented changes specified in #155 

## Verification

There were 4 tests that broke with this change because they were explicitly expecting `onEnd` to come before `onStatus`, so I updated those. I also wrote 4 new tests to ensure that the `onEnd` handler is called with a status and that that status matches the one returned from `onStatus`.


jonnyreeves:
Published to npm as 0.9.1-pre.c8a573d

PR #163, opened 2019-04-08
bugfix/162 - add cloneMessage to typescript definition files

jcollins-axway:
## Changes
- Added 1 line to src/ts/message that adds a cloneMessage call, using the Message name as the return type

## Verification
- Ran lint, test, and generate against the updated sources
- verified that the examples, also part of the commit, included the cloneMessage definition

### DIFF
--- a/examples/generated/proto/examplecom/annotations_pb.d.ts
+++ b/examples/generated/proto/examplecom/annotations_pb.d.ts
@@ -20,6 +20,7 @@ export class AnnotatedMessage extends jspb.Message {
   setMysfixed64(value: string): void;
 
   serializeBinary(): Uint8Array;
+  cloneMessage(): AnnotatedMessage;
   toObject(includeInstance?: boolean): AnnotatedMessage.AsObject;
   static toObject(includeInstance: boolean, msg: AnnotatedMessage): AnnotatedMessage.AsObject;
   static extensions: {[key: number]: jspb.ExtensionFieldInfo<jspb.Message>};
--- a/examples/generated/proto/examplecom/empty_message_no_service_pb.d.ts
+++ b/examples/generated/proto/examplecom/empty_message_no_service_pb.d.ts
@@ -5,6 +5,7 @@ import * as jspb from "google-protobuf";
 
 export class NoService extends jspb.Message {
   serializeBinary(): Uint8Array;
+  cloneMessage(): NoService;
   toObject(includeInstance?: boolean): NoService.AsObject;
   static toObject(includeInstance: boolean, msg: NoService): NoService.AsObject;
   static extensions: {[key: number]: jspb.ExtensionFieldInfo<jspb.Message>};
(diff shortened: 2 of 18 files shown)


jonnyreeves:
Thanks for raising this, however taking a quick look at the generated code checked into the repo, I can't see a `cloneMessage()` method on any of the generated classes (eg: [examplecom/annotations_pb.js](https://github.com/improbable-eng/ts-protoc-gen/blob/master/examples/generated/proto/examplecom/annotations_pb.js).

Was `cloneMessage()` introduced in a recent version of `protoc` (we use 3.5.1)? I did a quick google, but couldn't find anything.

jcollins-axway:
Hi Jonny,

It is documented here, [https://developers.google.com/protocol-buffers/docs/reference/javascript-generated#message](https://developers.google.com/protocol-buffers/docs/reference/javascript-generated#message)

In the google-protobuf library it is part of the Message class.

-Jason

jonnyreeves:
Thanks, but do you know which version of protoc it was released in, as the
generated code checked into this repo doesn't appear to contain a
`cloneMessage` function at first glance (although it's quite I'm missing
something and it's buried in a base-class?)

jcollins-axway:
That is exactly it.  It is part of the jspb.Message base class. 

[https://github.com/protocolbuffers/protobuf/blob/master/js/message.js](https://github.com/protocolbuffers/protobuf/blob/master/js/message.js)

jonnyreeves:
Awesome, thanks for clarifying :)

Sorry if this is a silly question, but doesn't that mean the generated
could should extend a base class (`Message`?), rather than in-lining the
`cloneMessage()` into the implementation?

jcollins-axway:
You are totally correct.  The generated message does extend the Message base.

However, when you use the cloneMessage function on a generated message class you then have to cast it back to the correct message class to avoid TS errors.

```typescript
let myMessage = new MyMessage();
myMessage.setName('name');

let myNewMessage  = myMessage.cloneMessage() as MyMessage;
myNewMessage.setName('name');

let myNewMessage2 = myMessage.cloneMessage();  //<-- no cast call
myNewMessage2.setName('name');  //<-- TS error: property 'setName' does not exist on type 'Message'
```

jonnyreeves:
Perhaps we could solve this problem by modifying the type definition for
the `Message` base class to make it generic, ie:

```
class Message<T extends Message<any>> {
  /* ... */
 cloneMessage(): T {
   /* ... */
 }
}
```

Personally, I would prefer this as it more accurately conveys what is
happening in the JavaScript code - what do you think?

jonnyreeves:
Alternatively, as a smaller change, you could just make `cloneMessage`
generic (source:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google-protobuf/index.d.ts#L112
)

jcollins-axway:
I see what you are saying John.  Both of those are better solutions to solve this.

jonnyreeves:
Please reach out if you need any help with the implementation; PRs against
the definitelytyped repo are usually merged pretty quickly :)

jcollins-axway:
Thank you for your help John!

jonny-improbable:
Hey @jcollins-axway do you have an update on this issue? Anything I can help with?

jcollins-axway:
Hey @jonny-improbable thanks for the ping I got caught up in some tasks at work.  Picking this up today.

jonny-improbable:
@jcollins-axway ping :) Anything I can help with?

jcollins-axway:
A maintainer on the DefinitelyTyped repo rejected the PR saying that it is not needed as we can cast the return to the Object we want.  I commented back that we should not need to cast it back as the method lives on the object calling it but received no response.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35602

Issue #166, opened 2019-04-11
MacOS: Sed errors when building on mac


strican:
**Versions of relevant software used**

macOS 10.14.4
Bazel 0.22.0
Building from [commit cdb79ec](https://github.com/improbable-eng/ts-protoc-gen/commit/cdb79ecf85a0ea08b86cf98860c2d42ffcc18d20). 

**What happened**

Building a `typescript_proto_library` fails on MacOS. Relevant output below in logs.

**What you expected to happen**

Build successfully.

**How to reproduce it (as minimally and precisely as possible)**:

On MacOS, clone the latest version of master.
Run `bazel build ...`

**Full logs to relevant components**

<details>Logs
<p>

```Bash
ts-protoc-gen (mac-sed-errors)$ bazel build ...
INFO: Invocation ID: 99cd2517-8b10-4318-8fb6-c705b3531199
DEBUG: /private/var/tmp/_bazel_steventricanowicz/9873f923b91e630355f329e396f67465/external/bazel_skylib/skylark_library.bzl:1:1: WARNING: skylark_library.bzl is deprecated and will go away i
n the future, please use bzl_library.bzl instead.
INFO: Analysed 24 targets (7 packages loaded, 112 targets configured).
INFO: Found 24 targets...
ERROR: /Users/steventricanowicz/source/ts-protoc-gen/proto/othercom/BUILD.bazel:9:1: error executing shell command: '/bin/bash -c bazel-out/host/bin/external/com_google_protobuf/protoc --plu
gin=protoc-gen-ts=bazel-out/host/bin/external/ts_protoc_gen/bin/protoc-gen-ts --ts_out=service=true:bazel-out/darwin-fastbui...' failed (Exit 1) bash failed: error executing command /bin/bas
h -c ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
sed: 1: "bazel-out/darwin-fastbu ...": undefined label 'azel-out/darwin-fastbuild/bin/proto/othercom/external_child_message_pb.js'
INFO: Elapsed time: 37.497s, Critical Path: 31.88s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

</p>
</details>

**Anything else we need to know**

[...]

PR #167, opened 2019-04-11
Splitting out JS code generation into its own rule

strican:
## Changes

In order to unblock development on Mac, I'm splitting out the Javascript code generation into its own rule. This seems like a generally good thing, and it allows for another rule for AMD module generation to cleanly be added on.

## Verification

Built (after locally fixing the sed issues described in #166) and it succeeded. Building on Mac excluding the new `javascript_library` rules also works.

Ran `yarn test` successfully.

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -4,10 +4,15 @@ load("@build_bazel_rules_nodejs//:defs.bzl", "yarn_install")
 
 TypescriptProtoLibraryAspect = provider(
     fields = {
-        "js_outputs": "The JS output files produced directly from the src protos",
         "dts_outputs": "Ths TS definition files produced directly from the src protos",
-        "deps_js": "The transitive JS dependencies",
         "deps_dts": "The transitive dependencies' TS definitions",
+    },
+)
+
+JavascriptProtoLibraryAspect = provider(
+    fields = {
+        "js_outputs": "The JS output files produced directly from the src protos",
+        "deps_js": "The transitive JS dependencies",
     },
 )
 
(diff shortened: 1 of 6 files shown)


strican:
Requesting review from @Dig-Doug, @jonnyreeves

strican:
There are two alternative approaches I can take to this:

1. Leave the JS generation as a part of the `typescript_proto_library` rule, but remove the AMD module generation. I didn't do this because I wasn't sure whether or not it was necessary. And either way, if we deemed it worth including as a default the first time, it seems like a bad idea to roll it back and break that promise for the JS code.
2. Leave it as is and just remove the AMD module generation for now. This is by far the simplest but (in addition to the concerns above) I worried about consumers building on Linux having no way to get access to the functionality that may have been expected.

strican:
Also, I'm going to work on some other things until I get at least a review of my approach. I'll address the CI failures after I've gotten confirmation that this is the way I should be doing this.

Dig-Doug:
Hey @strican,

Thanks for the PR. However, I don't think these changes have the effect you want. But first, let my answer your questions from the issue:

-----

### Why do we need js files?

This project, ts-protoc-gen, only generates typescript _declaration_ files. These files don't actually have the implementation of the protocol buffer inside of them, they only contain the list of methods that the proto has. This is used by the TS compiler to check types in typescript. When you compile typescript code, it's converted into JS. When you run it, you need to have the js files with the proto logic.

### What's with the AMD module generation?

I've been using the [angular-bazel-example](https://github.com/angular/angular-bazel-example) and the [rules_typescript code now in rules_nodejs](https://github.com/bazelbuild/rules_nodejs/blob/master/packages/typescript/internal/protobufjs/ts_proto_library.bzl) as the canonical examples for using bazel + typescript. In these examples, they use `requirejs` to load code in the browser.

If you look at the original generated JS code, it looks like this:

```js
/**
 * @fileoverview
 * @enhanceable
 * @suppress {messageConventions} JS Compiler reports an error if a variable or
 *     field starts with 'MSG_' and isn't a translatable message.
 * @public
 */
// GENERATED CODE -- DO NOT EDIT!

var jspb = require('google-protobuf');
```

This format is called `commonjs`, and it can't be [loaded by requirejs](https://requirejs.org/docs/commonjs.html). However, it can be easily converted manually (see the link for more info) into a AMD module. The code I wrote does some additional work to make it a named AMD module which is needed because proto files depend on each other.

NOTE: Other tools, e.g. webpack, can do this conversion for you in their build process. I saw that the rules_typescript also generates AMD modules by hand, so I based my implementation off of that.

### Why not split out the AMD generation into another rule?

From the user's point of view, that doesn't really make sense. As a user, I just want to list my proto file and get a fully working typescript library that I can list as a dep in other rules. I shouldn't need to also create a js_proto_lib, put the outputs of that into a js_module_lib, and then into a ts_lib.

-----

OK, so back to this PR. While your change will compile just fine, it won't work at runtime because the requirejs loaded won't be able to load the JS code properly. If you try running the test that failed in the ci:

```bash
bazel run //test/bazel:pizza_service_proto_test_suite_chromium-local
```

and look at the console, you'll see a bunch of loading errors.

As I mentioned above, I don't think separating the JS code is the right approach. If you're looking for a quick fix, I'd suggest editing the `sed` commands so that they work on OSX. The best long term solution I've come up with is still moving that code to nodejs, since I don't think there's any regex tools available in bazel actions. I think platform specific bash scripts will still run into the same issue because it's possible some machines have different versions of sed, etc.

strican:
Thanks for the write-up. I knew I didn't have enough context on this. :(

I'm gonna close this review and move the conversation back to the issue.

Issue #169, opened 2019-04-15
typescript_proto_library fails when depending on an external proto


strican:
**Versions of relevant software used**
master - commit [cbe73984](https://github.com/improbable-eng/ts-protoc-gen/commit/cbe73984ea86f4fde05722849811cff189db822a)
bazel 0.22.0
macOS 10.14.4

**What happened**
Building a `typescript_proto_library` when the proto is imported from an external dependency fails. This is because the output path for `protoc` is hard-coded based on the label while `declare_file` knows its running in the aspect context.

**What you expected to happen**
Build is successful and artifacts are correctly binplaced into the external namespace.

**How to reproduce it (as minimally and precisely as possible)**:
In one workspace, create a `proto_library`.
In a separate workspace, create a single BUILD file that depends on that proto:

```
typescript_proto_library(
  name = "external_ts_proto",
  proto = "@external//path/to:external_proto",
  visibility = ["//visibility:public"],
)
```

strican:
This is still valid and is addressed by #170. I'll update the PR with the latest in master. Not sure if I'll be able to get much time to finish it, but I think it's still worth keeping open.

PR #170, opened 2019-04-15
Adding a method to determine the correct output directory

strican:
## Changes

The change adds a method to dynamically determine the proper output directory for the protoc command. Currently, it assumes the bin directory. However, because the aspect may run in the context of an external project, that isn't sufficient. The behavior of `declare_file` properly handles this, so we use the directory of an output file to figure out what the correct output directory should be. 

## Verification

Verified by building
1. This repo and running `yarn test`.
2. A repo that builds protos and typescript_proto_libraries locally.
3. Another repo that builds typescript_proto_libraries that depend on protos from the first repo.

All are correctly placed in `bazel_out/<arch>/bin/path/to/proto` or `bazel_out/<arch>/bin/external/workspace/path/to/proto`, respectively.


strican:
Requesting review from @Dig-Doug, @jonny-improbable, @RNabel

defs.bzl
```diff
@@ -74,6 +74,19 @@ def _convert_js_files_to_amd_modules(ctx, js_protoc_outputs):
...
 
+def get_output_dir(ctx, outputs):
+    # Default to using bin_dir
+    output_dir = ctx.bin_dir.path
+
+    if len(outputs) <= 0:
+        return output_dir
+
+    declared_output = outputs[0]
+    if declared_output.dirname.endswith(ctx.label.package):
```
RNabel: Could you add a comment here to describe what this check is doing, maybe by documenting which path formats are expected to be seen here?

defs.bzl
```diff
@@ -75,15 +75,45 @@ def _convert_js_files_to_amd_modules(ctx, js_protoc_outputs):
...
+
+    # This check just verifies that the end of the declared output matches
+    # the package label. This should be true for all outputs so we just take
+    # the first.
     declared_output = outputs[0]
     if declared_output.dirname.endswith(ctx.label.package):
         output_dir = declared_output.dirname[:len(declared_output.dirname) - len(ctx.label.package)]
+    else:
+        print("""The declared outputs do not match their package labels. If you see errors, 
+                check to make sure your package labels match your directory structure.""")
```
RNabel: Package labels are defined by the directory structure, so I don't think it is possible for this `else` to ever be hit. Did you run into this case? 

From the [docs](https://docs.bazel.build/versions/master/build-ref.html#packages):

> A package is defined as a directory containing a file named BUILD or BUILD.bazel

strican: No, I did not, but I wasn't 100% sure about the structure of other projects so I added it out of an abundance of caution. I figured in the worst case it never gets hit and in the best case a user gets an actionable debug message. Do you think I should get rid of it?

Dig-Doug:
Just to make sure I understand: are you trying to do something like this?
```
proto_library(
  name = "test_proto",
  srcs = ["test.proto"],
  deps = [
    "@com_google_protobuf//:timestamp_proto",
  ],
)

# define test_ts_proto...
```
Can you add a test that reproduces the problem you're fixing?

defs.bzl
RNabel: Actually, given that there _may_ be future changes that could subtlely change how output files are declared, it's fine to keep this. I'd recommend adding the actual and expected output directory to the error message though.

strican:
@Dig-Doug, actually no, that's not the case I had. The case I was hitting was

```
typescript_proto_library(
  name = "test_ts_proto",
  proto = "@protos//api:test_proto",
  visibility = ["//visibility:public"],
)
```

I can see about setting up a test. Seems like I should be able to make a `typescript_proto_library` from some of the google protobufs.

strican:
Ping: @RNabel @Dig-Doug @jonny-improbable

test/bazel/proto/BUILD.bazel
```diff
@@ -17,3 +17,8 @@ typescript_proto_library(
     name = "pizza_service_ts_proto",
     proto = ":pizza_service_proto",
 )
+
+typescript_proto_library(
+    name = "external_ts_proto",
+    proto = "@external_protos//:external_proto",
```
Dig-Doug: Why not use a well known proto, e.g. timestamp_proto? You wouldn't have to create a whole new repo then.

test/bazel/proto/BUILD.bazel
```diff
@@ -17,3 +17,8 @@ typescript_proto_library(
     name = "pizza_service_ts_proto",
     proto = ":pizza_service_proto",
 )
+
+typescript_proto_library(
```
Dig-Doug: This isn't really a test since right now the only thing that's checked in the CI is if this rule builds. I guess it's OK, but it would be better to use this proto in a web test or check that the file is where you expect it to be (e.g. in an sh_test).

test/bazel/proto/BUILD.bazel
strican: I think I tried doing that but had trouble referencing them. I might not have tried timestamp though, and was looking for a shared proto in an already imported project. I can try with timestamp.

test/bazel/proto/BUILD.bazel
strican: Can you be more specific? I don't see the sh_tests you're referring to. And are you suggesting writing a test like [`pizza_service_proto_test.spec.ts`](https://github.com/improbable-eng/ts-protoc-gen/blob/master/test/bazel/pizza_service_proto_test.spec.ts)?

Dig-Doug: My bad, let me say it a different way: It would be better to use the generated file explicitly in a test. Two ways you could do this are (I'm suggesting do one):

1. Use the file in a new test, just like `pizza_service_proto_test.spec.ts`
2. Write a test that reads the file and checks the contents. This could be implemented in many ways, an [sh_test](https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/BUILD#L66) is just an example.

#-1 is probably easier since there are existing tests you can base the test off of.

test/bazel/proto/BUILD.bazel
strican: This is the error I get when I try to swap out the external_proto for one of the protobuf ones:

```Bash
ERROR: file 'external/com_google_protobuf/type_pb_amd.js' is generated by these conflicting actions:
Label: @com_google_protobuf//:type_proto
RuleClass: proto_library rule
Configuration: b8c06cd9b0b7503287e86fef3f8d5e1f
Mnemonic: SkylarkAction
Action key: 56265e69de57015eea1415766535c1c8, f961eaf8efde02547c4e263567c12a69
Progress message: Creating AMD module for @com_google_protobuf//:type_proto
PrimaryInput: File:[[<execution_root>]bazel-out/darwin-fastbuild/bin]external/com_google_protobuf/type_pb.js
PrimaryOutput: File:[[<execution_root>]bazel-out/darwin-fastbuild/bin]external/com_google_protobuf/type_pb_amd.js
Primary outputs are different: 1005642992, 148794312
Owner information: []#//:defs.bzl%typescript_proto_library_aspect BuildConfigurationValue.Key[b8c06cd9b0b7503287e86fef3f8d5e1f] @com_google_protobuf//:type_proto BuildConfigurationValue.Key[b8c06cd9b0b7503287e86fef3f8d5e1f] false {}, []#@ts_protoc_gen//:defs.bzl%typescript_proto_library_aspect BuildConfigurationValue.Key[b8c06cd9b0b7503287e86fef3f8d5e1f] @com_google_protobuf//:type_proto BuildConfigurationValue.Key[b8c06cd9b0b7503287e86fef3f8d5e1f] false {}
MandatoryInputs: Attempted action contains artifacts not in previous action (first 5): 
        external/com_google_protobuf/type_pb.js
        src/bazel/convert_to_amd_bin.sh
        src/bazel/convert_to_amd
        _middlemen/src_Sbazel_Sconvert_Uto_Uamd-runfiles
Previous action contains artifacts not in attempted action (first 5): 
        external/com_google_protobuf/type_pb.js
        external/ts_protoc_gen/src/bazel/convert_to_amd_bin.sh
        external/ts_protoc_gen/src/bazel/convert_to_amd
        _middlemen/external_Sts_Uprotoc_Ugen_Ssrc_Sbazel_Sconvert_Uto_Uamd-runfiles
Outputs: Attempted action contains artifacts not in previous action (first 5): 
        external/com_google_protobuf/type_pb_amd.js
Previous action contains artifacts not in attempted action (first 5): 
        external/com_google_protobuf/type_pb_amd.js
ERROR: com.google.devtools.build.lib.actions.MutableActionGraph$ActionConflictException: for external/com_google_protobuf/type_pb_amd.js, previous action: action 'Creating AMD module for @co
m_google_protobuf//:type_proto', attempted action: action 'Creating AMD module for @com_google_protobuf//:type_proto'
INFO: Elapsed time: 0.333s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 7 targets configured)
```

Dig-Doug: I tried cloning you repo -- that error is caused by my previous change, the tool I added isn't referenced properly to work in external repos. In `defs.bzl`, it should be:

`//src/bazel:convert_to_amd` -> `"@ts_protoc_gen//src/bazel:convert_to_amd`

After changing that, I get another error about the files not being created. It looks like the external generated proto files are not being written to the correct directory. I ran out of time to debug tonight, but you can run: `bazel build -s --verbose_failures --sandbox_debug //test/bazel/proto:external_ts_proto` to see the command and look at the file structure.

strican: Ohhhh boy. Thanks for this. This makes it clear what the problem is. The files are being output to `bazel-bin/external/com_google_protobuf/google/protobuf/` but we're expecting them to be output to `bazel-bin/external/com_google_protobuf/`. That's because the _bazel_ package name is empty (`//:timestamp_proto`) but the declared package is [`google.protobuf`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto). So ironically, I was correct to put a check in to ensure the packages match. However, it's not the bazel package, it's the declared proto package, which I'm not sure if we can actually get access to in any way from this aspect (seems related to the same issue I'm running into [with Go](https://github.com/bazelbuild/rules_go/issues/707)). Plus, the check doesn't work when the bazel package is empty, so there's that, too...

Thoughts on how to resolve this?

defs.bzl
```diff
@@ -104,7 +147,7 @@ def typescript_proto_library_aspect_(target, ctx):
```
strican: So apparently this doesn't behave as I thought it did. I expected the aspect to use the dependency graph to build the correct outputs. However, this doesn't actually declare the outputs for any of the dependencies. It seems like the for loop should be:

`for src in depset(target.proto.direct_sources) + target.proto.transitive_sources:`

So that it actually builds the union of all of the direct sources and any dependencies. Otherwise, you'll get import references to files that don't end up getting copied over to bazel-bin. @Dig-Doug, did you intend for this?

Note - this is another way to expose the problem I encountered with com_google_protobuf. As soon as one of the transitive sources is in a different package, it will break. I'm wondering if what I'm trying to do here is even possible. I was under the impression that it was, but now I'm not sure how to accomplish it.

test/bazel/proto/BUILD.bazel
Dig-Doug: I took another look at this this morning and I think I've got it working.

There are two problems:

- Output files are not declared in the right directory
- The protoc output dir is not correct

For #-1, external files were being delcared as `external/com_google_protobuf/proto.proto` instead of `external/com_google_protobuf/google/protobuf/proto.proto`.

For #-2, the output dir should be the root of the repo, not the directory of the file. protoc handles writing to the sub dirs.

This reminded me of a similar [issue](https://github.com/pubref/rules_protobuf/pull/93) I fixed a few years ago. It may make sense to move this code into the rules_proto project so that we don't have to reinvent the wheel -- that's separate from this change though.

[Here are my changes](https://gist.github.com/Dig-Doug/5c971c2b3fe65046aabcc214677dbe9d), the output is a git diff. Can you see if they work for your use case? If so, feel free to patch them into this PR.

strican: Interesting. I was under the impression that protoc would output the files based on the package name, but it looks like it's done based on the path of the proto file. I'm not sure if that's desirable in the long-run, but it definitely makes this easier now. I see your solution makes sense now though - it just took me a while to figure out that it would generalize.

defs.bzl
Dig-Doug: Missed this before. You're correct that we want to generate the files for the transitive dependencies. This way this works is the aspect is run by bazel on each node in the dependency graph. What I do is I pass the generated outputs up from the direct dependencies [here](https://github.com/improbable-eng/ts-protoc-gen/blob/master/defs.bzl#L130), so all the transitive dependencies bubble up to the root (in `dep_js`).

Is the problem you mentioned fixed now?

test/bazel/proto/BUILD.bazel
Dig-Doug: Can you add a similar ts proto lib for a google protobuf and use both of these new libs in a karma web test? [example](https://github.com/improbable-eng/ts-protoc-gen/blob/master/test/bazel/BUILD.bazel#L55)

strican: @Dig-Doug - I see you importing from an ambient module, but I don't see how/where those are defined. I tried using what I saw in the *_amd.js file, but that didn't seem to work. For instance, writing a test for this, I used `ts_protoc_gen/external_pb` but would get the error `Cannot find module 'ts_protoc_gen/external_pb'`

strican: So I managed to get it compiling by using a relative module. I think we still want to eventually switch it back though.

That said, when I run it as I have it now, I get this (rather unscrutable) error:

```
[33m29 04 2019 18:53:55.784:WARN [web-server]: [39m404: /ts_protoc_gen/external/external_protos/external_pb.js
HeadlessChrome 69.0.3497 (Mac OS X 10.14.4) ERROR: 'There is no timestamp for ts_protoc_gen/external/external_protos/external_pb.js!'

HeadlessChrome 69.0.3497 (Mac OS X 10.14.4) ERROR
  {
    "message": "Uncaught Error: Script error for \"ts_protoc_gen/external/external_protos/external_pb\", needed by: ts_protoc_gen/test/bazel/external_proto_with_deps_test.spec\nhttp://requirejs.org/docs/errors.html#scripterror\nat npm/node_modules/requirejs/require.js:143:9\n\nError: Script error for \"ts_protoc_gen/external/external_protos/external_pb\", needed by: ts_protoc_gen/test/bazel/external_proto_with_deps_test.spec\nhttp://requirejs.org/docs/errors.html#scripterror\n    at makeError (node_modules/requirejs/require.js:168:17)\n    at HTMLScriptElement.onScriptError (node_modules/requirejs/require.js:1738:36)",
    "str": "Uncaught Error: Script error for \"ts_protoc_gen/external/external_protos/external_pb\", needed by: ts_protoc_gen/test/bazel/external_proto_with_deps_test.spec\nhttp://requirejs.org/docs/errors.html#scripterror\nat npm/node_modules/requirejs/require.js:143:9\n\nError: Script error for \"ts_protoc_gen/external/external_protos/external_pb\", needed by: ts_protoc_gen/test/bazel/external_proto_with_deps_test.spec\nhttp://requirejs.org/docs/errors.html#scripterror\n    at makeError (node_modules/requirejs/require.js:168:17)\n    at HTMLScriptElement.onScriptError (node_modules/requirejs/require.js:1738:36)"
  }
```

Here are my bazel targets:

```Python
ts_library(
    name = "external_proto_with_deps_test",
    testonly = 1,
    srcs = ["external_proto_with_deps_test.spec.ts"],
    tsconfig = ":tsconfig.json",
    deps = [
        "//test/bazel/proto:external_ts_proto",
        "@npm//@types/jasmine",
        # Don't put external_dep_ts_proto here, we want to test it is included transitively
    ],
)

ts_web_test_suite(
    name = "external_proto_with_deps_test_suite",
    browsers = [
        "@io_bazel_rules_webtesting//browsers:chromium-local",
    ],
    deps = [
        ":google_protobuf_requirejs",
        ":external_proto_with_deps_test",
    ],
)
```

strican: So I'm wondering if the issue is that `external_pb_amd.js` specifies `define("ts_protoc_gen/external_pb", ...` for the module but the error output indicates it expects `ts_protoc_gen/external/external_protos/external_pb`. Indeed, when I use that as the import, it does compile, so that seems to be the registered ambient module, but I'm still getting errors at runtime.

Dig-Doug: Yes, it looks like there is a problem with the external AMD modules names. I think it should be something like `external_protos/external_pb`. The format is `<workspace-name>/path/to/proto`, e.g. `ts_protoc_gen/test/bazel/proto/pizza_service_pb`.

strican: Okay, so that sounds like I should probably factor out some of this logic and change the arguments we're calling `ctx.executable._convert_to_amd` with.

Will that solve the problem completely though? If it builds, it's clearly finding the module, even if it's got the wrong name. What are the runtime errors I'm getting?

strican: This is starting to seem crazy, too. We can't be the only aspect that has this problem. It feels like stuff like `isExternal` and `workspace_name` should be available per dependency and not only from the original context. Is this a core bazel problem?

Dig-Doug: Actually I was just thinking about this a bit more, one more different thought:

In the proto files, you don't specify the workspace when you import them. If you look at the generate js code from protoc, it also isn't including the workspace when looking for the modules. So what I'm thinking is that the AMD module names shouldn't include the workspace. (The file locations are fine I think, it just the names inside the files that need to change). What do you think?

Not sure if this will fix everything, haven't tried it locally.

Dig-Doug: Actually I was just thinking about this a bit more, one more different thought:

In the proto files, you don't specify the workspace when you import them. If you look at the generate js code from protoc, it also isn't including the workspace when looking for the modules. So what I'm thinking is that the AMD module names shouldn't include the workspace. (The file locations are fine I think, it just the names inside the files that need to change). What do you think?

Not sure if this will fix everything, haven't tried it locally.

strican: I guess what happens if I have two dependencies from two different external workspaces that have the same name/location? There'd be no way to uniquely identify them. This isn't an issue I'm having (only importing from a single workspace atm) but it feels like workspace has to be necessary.

strican: Omg, this stuff is there and we're doing it in `proto_path`. Okay, that gives me a lot to go off of. I think I might be able to clean this up.

strican: I got it so that if you look at `external_pb_amd.js` the module is defined as `external/external_protos/external_pb` but I still get a compile error if I import that. The error output indicates it's the same as above: `ts_protoc_gen/external/external_protos/external_dep_pb`. Clearly, it's getting it from somewhere other than the _amd.js file, so any ideas on where that would be? I'm assuming it's based on the directory structure, but since it's compile time, the only way to get an ambient module is to import it via node, right?

Like I said earlier, I'm back to the part of the problem where it compiles and it still fails.

Dig-Doug: I did some debugging, here's what I've found:

- Currently, I believe the only way to import code from an external repo in TS is to use a path like `workspace/external/external_repo/path`, e.g. `ts_protoc_gen/external/external_protos`. This looks like a bug to me in rules_typescript, (I think just `external_protos/` should be valid), but I'm not sure -- I'll create a reproduction and file an issue with them.

- When the typescript code is translated into JS, it uses the import path as the amd module name.  So, the AMD module name of the proto must be `ts_protoc_gen/external/external_protos/external_pb`.

For now, I think we should just make the proto AMD module name have the workspace name, `ts_protoc_gen/`, at the beginning. If in the future the external repo path is fixed, we can change the code to export AMD modules like `external_protos/external_pb`. What do you think?

kellycampbell: For #1 you can just add `sibling=src` in `append_to_outputs` where it does `declare_file`, e.g.

```
output = ctx.actions.declare_file(file_name + ext, sibling=src)
```

strican:
Updating the PR with the latest from master to make sure #169 isn't marked as stale.

PR #178, opened 2019-05-27
Support bazel external protos such as google well known types

kellycampbell:
## Changes

- when using bazel rules, support "external" proto libs

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -29,11 +29,11 @@ def proto_path(proto):
         path = path[1:]
     return path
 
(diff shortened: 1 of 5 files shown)


kellycampbell:
Having some trouble getting the unit test to find the timestamp_pb.js file. I currently only use this on the server-side with nodejs. 

Example project here: https://github.com/kellycampbell/example-bazel-proto-ts

kellycampbell:
Abandoning this because it only handles well-known protos and #170 is further along at supporting all external deps.

Issue #177, opened 2019-05-22
Add a bazel flag for enabling/disabling conversion from CommonJS to AMD step.


pbsf:
I think it would be interesting to have a flag for enabling/disabling the conversion from the generated CommonJS code to AMD. On my project I'm currently using a copy of the `defs.bzl` file, removing the call to _convert_js_files_to_amd_modules for this plugin to work for a React Native scenario.

Thoughts?

mattem:
I'm after something similar, as we need to take the generated protos into `rollup_bundle`. 
I don't think a flag is the correct approach here though, as it's normally down to the downstream rule to request the right output (AMD, es6 etc) rather than the user passing in a flag. 

I've been experimenting producing both es5 and es6 sources, instead of setting the es5 output to both:

```
es5_sources = es5_js_outputs + aspect_data.deps_js,
es6_sources = es5_js_outputs + aspect_data.deps_js,
```

Dig-Doug:
I ran into the `rollup_bundle` problem as well today so took a stab at fixing the es6 output in #185.

PR #179, opened 2019-05-27
Add module param to bazel rules for specifying amd or commonjs

kellycampbell:
## Changes
   - add option to bazel typescript_proto_library to leave the generated code in commonjs format for nodejs use

## Verification
  - Example project at https://github.com/kellycampbell/example-bazel-proto-ts

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -123,7 +123,10 @@ def typescript_proto_library_aspect_(target, ctx):
     )
 
     dts_outputs = depset(dts_outputs)
-    js_outputs = depset(_convert_js_files_to_amd_modules(ctx, js_protoc_outputs))
+    if ctx.attr.module == "amd":
+        js_outputs = depset(_convert_js_files_to_amd_modules(ctx, js_protoc_outputs))
+    else:
+        js_outputs = depset(js_protoc_outputs)
     deps_js = depset([])
     deps_dts = depset([])
 
@@ -143,6 +146,7 @@ typescript_proto_library_aspect = aspect(
     implementation = typescript_proto_library_aspect_,
     attr_aspects = ["deps"],
     attrs = {
+        "module": attr.string(values = ["amd", "commonjs"]),
         "_ts_protoc_gen": attr.label(
             allow_files = True,
             executable = True,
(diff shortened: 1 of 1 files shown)


kellycampbell:
Fixes #177

jonny-improbable:
@Dig-Doug, could you please take a look?

Dig-Doug:
@kellycampbell, thanks for the PR.

I originally considered this approach when I was implementing the logic for this rule. As I was looking through the existing JS proto library implementations, I came across [this comment](https://github.com/grpc/grpc-web/pull/260#issuecomment-415443448). It explains why having a user-defined import style is problematic in some cases.

After looking at #177, I agree with @mattem, it would be better if the output type could be requested. Currently ES6 code is not being generated, have you tried updating the rule to generate it separately? Would that work for your use case?

kellycampbell:
@Dig-Doug thanks. That makes sense. My use-case is server-side nodejs. I think the es6 method would work for that. @mattem do you already have something in progress for converting the protoc output files to es6?

eordano:
Hi there,

First of all, thank you very much for maintaining this library. I'm using bazel 0.28 and found @Dig-Doug's work very helpful (and recent)

Secondly, I feel that even though allowing users to specify the module mechanism might be problematic -- it's worse than the alternative (will have to fork).

eordano:
(I'm looking for `es6` imports -- amd doesn't work for us)

Dig-Doug:
Hi @eordano, do you have some more info about the problem your having?

I had a need for es6 too and made some changes in #185. Have you tried those? If you're still having issues with the standard TS / JS bazel toolchain, we should fix them.

P.S. there is a pending PR for compatibility with 0.28 :)

eordano:
Found #185 later thank you!

jonny-improbable:
@kellycampbell, @Dig-Doug what are the suggested next steps for this PR? It's currently in conflict so it needs rebasing at the very least.

Dig-Doug:
I actually tried using NodeJS yesterday myself and found a few issues :) -- I've submitted #194 to fix them.

@kellycampbell - Can you try out the changes in #194 and see if it works for you?

jonny-improbable:
@Dig-Doug is this pull now obsolete given #194 has been merged?

Dig-Doug:
Yes, I believe so. I've been using the TS protos in nodejs successfully.

eordano:
Can also confirm it's working 👍

PR #185, opened 2019-06-25
Bazel rules add ES6 output

### LINKED ISSUE #177: Add a bazel flag for enabling/disabling conversion from CommonJS to AMD step.

Dig-Doug:
ES5 srcs are unchanged, ES6 srcs are commonjs. I've used the `rollup_bundle` rule to test the ES6 srcs -- it will fail if they can't be resolved.

Fixes #177


mattem:
Hey @Dig-Doug, apologies I've been away on vacation for a while and missed the comments on the other PR.

Thanks for putting this together! I've tested this out against our apps with rollup and it seems good.

jonny-improbable:
Two thumbs up from the community and a green CI build - let's land this.

@Dig-Doug could you please update the Overview and provide me with a line or two that I can use in the CHANGELOG? Thanks! :)

Dig-Doug:
Done, thanks!

Issue #27, opened 2017-09-13
Can this be used to generate grpc service definitions?


danbopes:
In the [grpc node examples folder](https://github.com/grpc/grpc/blob/master/examples/node/static_codegen/README.md) they generate the contracts between the services, but also the service itself [_grpc_pb.js](https://github.com/grpc/grpc/blob/master/examples/node/static_codegen/helloworld_grpc_pb.js). I notice that you can add service=true to generate the grpc-web bindings, but what about the grpc service definitions?

MarcusLongmuir:
@danbopes: This project (ts-protoc-gen) currently only generates the method definitions in a format that https://github.com/improbable-eng/grpc-web requires. 

I think it might make sense to generate the node grpc declarations in a separate plugin - I realise by the same logic the grpc-web definitions should also be generated by a separate plugin, but there hasn't been a pressing need to.

To that end, creating a TypeScript library that can be used to build these various protoc plugins would be a good next step.

This could be done by splitting the parsing and printing steps that this plugin performs and creating an interface that can be used to print other formats.

jonny-improbable:
See #42

jonahbron:
@danbopes This is now available in version 0.6.0, this issue can be closed.

Issue #145, opened 2018-12-13
RFC: The future of this package


jonnyreeves:
## Overview
This package (ts-protoc-gen) has always conflated two distinct use-cases:

1. Generate TypeScript definitions (`.d.ts`) to accompany the JavaScript generated by `protoc`
2. Generate the JavaScript and TypeScript definitions for use with improbable-eng/grpc-web

In the light of recent forks, such as [adding support for flowtype](https://github.com/chrisgervang/flowtype-protoc-gen), I feel it may be wise to consider splitting these two use-cases out into two distinct protoc plugins.

1. `protoc-gen-tsd` (usage: `--tsd_out=$OUT`)
2. `protoc-gen-improbable-grpc-web` (usage: `--improbable-grpc-web_out=dts:$OUT`)

This renaming would also help create a distinction between the [official grpc-web offering](https://github.com/grpc/grpc-web) from google, and the [improbable-eng/grpc-web offering](https://github.com/improbable-eng/grpc-web) from Improbable.

## Migration Path
1. This repo (and associated npm package) could potentially continue to exist and wrap `protoc-gen-tsd` and `protoc-gen-improbable-grpc-web` to avoid breaking consumer workflows until such a point that they can be deprecated and migrated.
2. The code would need to be split; I can see two possible approaches:
  a.) Create 2 or more repositories under improbable-eng for each protoc plugin (and possibly one for shared code)
  b.) Create a single mono-repository using a tool like [lerna](https://github.com/lerna/lerna) to manage the node modules.

## Misc
It may make more sense to join efforts with [agreatfool/grpc_tools_node_protoc_ts](https://github.com/agreatfool/grpc_tools_node_protoc_ts) (cc @agreatfool) who has built out their own protoc compiler for generating TypeScript definitions. Although a full audit of feature-sets would need to follow.

cc @easyCZ @MarcusLongmuir @johanbrandhorst @chrisgervang

johanbrandhorst:
Great initiative, agreed that it will be a good thing for users to have more clarity on these two use cases.

Dig-Doug:
There are two main parts to the current bazel setup:

1. Rules needed to build the protoc plugins under bazel, e.g. [this target](https://github.com/improbable-eng/ts-protoc-gen/blob/master/bin/BUILD.bazel#L5)

[...]

PR #193, opened 2019-08-05
Add support for generating grpc-node service types

esilkensen:
## Changes

This PR adds support for generating [grpc-node](https://github.com/grpc/grpc-node) service types (see https://github.com/improbable-eng/ts-protoc-gen/pull/75, https://github.com/improbable-eng/ts-protoc-gen/issues/42, https://github.com/improbable-eng/ts-protoc-gen/issues/27):

* The existing `service=true` option has been replaced with `service=grpc-web`, following the suggestion in https://github.com/improbable-eng/ts-protoc-gen/pull/75#issuecomment-391747282.
* The new `service=grpc-node` option can be used to generate grpc-node service types, writing a `_grpc_pb.d.ts` file for each `_grpc_pb.js` file generated by the protoc-gen-grpc plugin.

## Verification

To facilitate testing of the `service` options and the new `service=grpc-node` option in particular, the generate.sh script has been modified following https://github.com/improbable-eng/ts-protoc-gen/pull/75#issuecomment-392140049:
* `examples/generated` contains the output when the `service` option is not used
* `examples/generated-grpc-web` contains the output for the `service=grpc-web` option
* `examples/generated-grpc-node` contains the output for the `service=grpc-node` option

The generate.sh script uses grpc-tools to generate `_grpc_pb.js` implementation files to help test that the generated `_grpc_pb.d.ts` files are consistent with them.


examples/generated-grpc-node/proto/examplecom/simple_service_grpc_pb.d.ts
```diff
@@ -0,0 +1,36 @@
...
+  doUnary: grpc.MethodDefinition<proto_examplecom_simple_service_pb.UnaryRequest, proto_othercom_external_child_message_pb.ExternalChildMessage>;
+  doServerStream: grpc.MethodDefinition<proto_examplecom_simple_service_pb.StreamRequest, proto_othercom_external_child_message_pb.ExternalChildMessage>;
+  doClientStream: grpc.MethodDefinition<proto_examplecom_simple_service_pb.StreamRequest, google_protobuf_empty_pb.Empty>;
+  doBidiStream: grpc.MethodDefinition<proto_examplecom_simple_service_pb.StreamRequest, proto_othercom_external_child_message_pb.ExternalChildMessage>;
+  delete: grpc.MethodDefinition<proto_examplecom_simple_service_pb.UnaryRequest, proto_examplecom_simple_service_pb.UnaryResponse>;
+}
+
+export const SimpleServiceService: ISimpleServiceService;
+
+export class SimpleServiceClient extends grpc.Client {
```
esilkensen: Here's an example of one of the generated grpc-node service types

package.json
```diff
@@ -46,13 +46,15 @@
...
+    "grpc": "^1.22.2",
+    "grpc-tools": "^1.8.0",
     "lodash": "^4.17.5",
     "lodash.isequal": "^4.5.0",
     "minimist": "^1.2.0",
     "mocha": "^5.2.0",
     "mocha-spec-json-output-reporter": "^1.1.6",
     "source-map-support": "^0.4.18",
-    "ts-node": "^5.0.1",
+    "ts-node": "^8.3.0",
```
esilkensen: As I was working on this, it seemed that ts-node 5.0.1 wasn't catching type errors in test/integration/service/grpcnode.ts - I tried updating to the latest 8.3.0 and then it did catch them and would cause `npm run test` to fail, as expected.

package.json
```diff
@@ -46,13 +46,15 @@
     "babel": "^6.5.2",
     "browser-headers": "^0.4.1",
     "chai": "^3.5.0",
+    "grpc": "^1.22.2",
```
esilkensen: Does this package need to be in "dependencies" since it is imported in the generated `_grpc_pb.d.ts` files, or is it ok here in "devDependencies"?

src/service/common.ts
```diff
@@ -0,0 +1,142 @@
...
+  readonly nameAsCamelCase: string,
+  readonly functionName: string,
+  readonly serviceName: string,
+  readonly requestStream: boolean
+  readonly responseStream: boolean
+  readonly requestType: string
+  readonly responseType: string
+};
+
+export class RPCDescriptor {
```
esilkensen: No changes to the code in this file, just extracted what was common between grpcnode.ts and grpcweb.ts

src/service/common.ts
```diff
@@ -0,0 +1,142 @@
...
+          requestStream: method.getClientStreaming(),
+          responseStream: method.getServerStreaming(),
+          requestType: callingTypes.requestType,
+          responseType: callingTypes.responseType,
+        };
+      });
+  }
+}
+
+export class GrpcServiceDescriptor {
```
esilkensen: This file is unchanged code extracted from grpcweb.ts - except for renaming this class from `GrpcWebServiceDescriptor` to `GrpcServiceDescriptor` since it wasn't necessarily specific to grpc-web.

test/mocha-run-suite.sh
```diff
@@ -19,7 +19,7 @@ fi
 mocha \
     --reporter mocha-spec-json-output-reporter \
     --reporter-options "fileName=./test/mocha-report.json" \
-    --require ts-node/register/type-check \
+    --require ts-node/register \
```
esilkensen: `type-check` is the default behavior in the newer ts-node

esilkensen:
Hi @jonny-improbable 👋

I wondered if this is still a feature you think is worth pursuing?

If so, I think this PR is potentially ready for a review. Whenever you have a chance to take a look, if there's anything I can do to help make it easier to review, I will be happy to do that.

(I wondered about breaking off an initial PR that did nothing but adding that `examples/grpc-web-services` directory, in the interest of shrinking the diff...)

Anyway, sorry to @ ping you - I'll look forward to any next steps to take with this PR, and of course will understand if this just isn't a path for this project to go down.

Cheers!

package.json
jonny-improbable: Yep should be fine, I see you are only consuming the `grpc` package in your tests (hence a dev dependency) so shouldn't cause an issue for consumers.

src/index.ts
```diff
@@ -39,10 +41,14 @@ withAllStdIn((inputBuff: Buffer) => {
       thisFile.setContent(printFileDescriptorTSD(fileNameToDescriptor[fileName], exportMap));
       codeGenResponse.addFile(thisFile);
 
-      if (generateServices) {
+      if (generateGrpcWebServices) {
         generateGrpcWebService(outputFileName, fileNameToDescriptor[fileName], exportMap)
           .forEach(file => codeGenResponse.addFile(file));
       }
+      if (generateGrpcNodeServices) {
```
jonny-improbable: This should be `else if` to reflect the fact that it's not possible to emit both `grpc-web` and `grpc-node` services at the same time.

test/integration/service/grpcweb.ts
```diff
@@ -44,31 +44,40 @@ describe("service/grpc-web", () => {
       assert.strictEqual(SimpleService.DoClientStream.responseType, Empty);
     });
 
-    it("should contain the expected  DoClientStream method", () => {
+    it("should contain the expected DoBidiStream method", () => {
```
jonny-improbable: 🙏

esilkensen:
Thanks Jonny!

1. _Migration Path_;  I pushed a change that preserves the behavior for `service=true`, but logs a warning that it has been deprecated and `service=grpc-web` should be used instead.

2. _Ownership_; I can definitely commit to this. I would be happy to jump on any issues, and also would be using this code personally at work.

3. _Placement_; Ah yes, I was looking at #145 and [grpc_tools_node_protoc_ts](https://github.com/agreatfool/grpc_tools_node_protoc_ts) (and [this fork](https://github.com/niklaus0823/protoc-gen-grpc-ts) of it too) as I was getting started on this PR. I think what you've laid out makes a lot of sense.

I ended up going down this path because:
* I really like the test coverage in ts-protoc-gen. grpc_tools_node_protoc_ts doesn't have any;  I think code was copied from ts-protoc-gen, but certain things like #156 haven't made it in.
* We're already using ts-protoc-gen, and recently had reason to start using grpc-node services. It'd be super easy to just bump a version number and pass a new `service` parameter.

Do you think there's a case for adding this functionality to ts-protoc-gen, for however much current users of the package may get out of it, with the understanding that you would still be pursuing #145 and these grpc-node services wouldn't be part of the plans going forward?

If so, I'd reiterate my commitment to owning any issues related to this code. But if not, I certainly understand, and could look at contributing and migrating to that other project instead.

(Thanks for the pointer to that ts-morph library, looks cool!)

jonny-improbable:
Morning @esilkensen,

Thanks for answers my questions above; it looks like we are all set! Sorry to be a pain, but I've just landed #195 which has caused a number of conflicts, in particular around the way the example protos are generated. Please could you resolve these and then we will be good to merge.

esilkensen:
Sounds good, thanks so much @jonny-improbable,

I've just resolved the conflicts and will watch to make sure CI passes.

(Then head to bed, it is _early_ morning for me 😴)

esilkensen:
Hi @jonny-improbable, I just wanted to check back in on this PR, to see if you were still thinking of merging it -- or if there are any additional changes I could make? Thanks again!

esilkensen:
Nice!

What do you think about versioning? Will this eventually make it into a 0.10.1 release? 

(I'll pull from `ts-protoc-gen@next` in the meantime)

jonny-improbable:
0.x versioning is a little hectic, I'll just bump this out at 0.11 as it
contains a major new feature.

PR #207, opened 2019-10-21
Bazel Deprecation

Dig-Doug:
Follow up to [this discussion](https://github.com/improbable-eng/ts-protoc-gen/issues/145#issuecomment-544270083).

This change does the following:
- Removes the bulk of bazel related code
- Leaves a empty implementation of the rules that prints a deprecation warning, in case someone tries to update their version without looking at the `README`
- Updates the `README` to point to the new implementation.

After this PR, the only remaining pieces of bazel related code are the `WORKSPACE`, the `BUILD`, and the `defs.bzl` files. They can be cleaned up sometime in the future.


Dig-Doug:
Fixes #171.

jonny-improbable:
Lovely stuff, thanks for your stewardship @Dig-Doug.

PR #212, opened 2019-11-06
WIP: Add generate parameter - Fixes #211

### LINKED ISSUE #211: Generate both grpc-node and grpc-web simultaneously
Would it make sense to be able to generate both grpc-node and grpc-web services at the same time, rather than the current if/else we have today?

Multiply:
## Changes
I've added a relatively crude parameter parse, that allows comma separated arguments, with multiple values delimited by `+`, as comma is used to distinguish between multiple parameters. 

## Verification
I've tested it by using a modified version of the bazel rules, following the comment here: https://github.com/Dig-Doug/rules_typescript_proto/issues/10#issuecomment-550334084
I couldn't seem to find any cli tests, so please guide me in the right direction, if I missed them.


jonny-improbable:
Thank you for raising this.

WRT code structure, can I suggest that you introduce a 'proper' type for the Parameters object instead of relying on `{[key: string]: string[]}` - this should make things a little more extensible/maintainable in future. I'd also suggest introducing a `parseParameters()` func to further improve readability.

Please could you also update the documentation? I would suggest that we replace the current references to `service=` with `generate=` and standardize on this new parameter.

Multiply:
We ended up not using this at all, after moving 100% to golang.

If anyone wants to pick it up, feel free to use my work here.

Issue #220, opened 2020-02-14
Generate Server bindings for grpc-node


Legogris:
It's possible that I am missing something, but it seems like the generated services only offer type checking for clients, not for servers.

It would be great if server implementations were also type-checked.

yukio808:
I have an example of this working that I will submit a pull request for. From the example will generate something like this:
```typescript
interface IOrphanServiceServer {
  doUnary: grpc.handleUnaryCall<proto_orphan_pb.OrphanUnaryRequest, proto_orphan_pb.OrphanMessage>;
  doStream: grpc.handleUnaryCall<proto_orphan_pb.OrphanStreamRequest, proto_orphan_pb.OrphanMessage>;
}
```

This is something that I need/*want* as well.

PR #221, opened 2020-02-24
GRPC-Node :: Generate Server Interface from Service Proto

yukio808:
## Changes
There was a request to add server definitions for NodeJS
Issue: [#220](https://github.com/improbable-eng/ts-protoc-gen/issues/220)
Created a server definition function that will print to file the Server Interface.

## Verification

1. Ran the Generate step in package.json
2. Checked example: orphan_grpc_pb.d.ts in examples/generated-grpc-node/proto


yukio808:
Stand By I think I might have a small bug.

PR #224, opened 2020-04-08
Include leading comments from proto file for message and enum types

rocwind:
Include the proto file leading comments for enum type and its values, message type and its `AsObject` type fields + getter methods

## Changes
* add `ts/CommentsMap` to parse the LocationList
* modify `ts/message`, `ts/enum` to write leading comments
* update generated example

## Verification
tested with the example proto


dhalperi:
This is great!

PR #247, opened 2020-09-23
Add Server interface

badsyntax:
Unless i'm completely missing something, it's not possible to use the types generated by this package for a server implementation. This change fixes that. 

I've had to make various upstream changes to allow the types to work across the `grpc` and `@grpc/grpc-js` packages. I've been following & contributing to the `proto-loader` typescript generator changes in https://github.com/grpc/grpc-node/pull/1474, and changes in this PR align with those changes. 

If using `@grpc/grpc-js`, a minimum version of `1.2.0` is required. I'm not sure how best to handle this, whether we would mention this in the readme, or add it as a `peerDependency`? 

- ~This PR is a draft because there's a [mismatch of types](https://github.com/grpc/grpc-node/issues/1581) between the `grpc` and `@grpc/grpc-js` packages. The code in this PR will generate types that only work with `@grpc/grpc-js`. I hope to get come clarity on the types mismatch soon.~
- ~UPDATE: Waiting on the [latest type changes to @grpc/grpc-js](https://github.com/grpc/grpc-node/pull/1587) to be published, to support this change.~
- ~UPDATE 2: We still need the types to match in `grpc` for this to work.~

## Changes

- Generate a server interface

## Example usage

```ts
import * as grpc from '@grpc/grpc-js';
import { ChatService, IChatServer } from './proto/chat_grpc_pb';
import { ClientMessage, ServerMessage } from './proto/chat_pb';

const chatServer: IChatServer = {
  join(call: grpc.ServerWritableStream<ClientMessage, ServerMessage>): void {
    // implementation
  },
  send(
    call: grpc.ServerUnaryCall<ClientMessage, ServerMessage>,
    callback: grpc.sendUnaryData<ServerMessage>
  ): void {
    // implementation
  },
};

const server = new grpc.Server();
server.addService(ChatService, chatServer);
```

## Verification

I've used the generated files in an example app and they seem ok. I briefly explored writing tests but all the existing tests are testing the javascript implementation and not the generated types, so not sure how best to create tests for typescript interfaces. 

I believe we can ignore the the failing build (linting) as this new code is consistent with existing code. 

## Related 

https://github.com/improbable-eng/ts-protoc-gen/pull/221


src/service/grpcnode.ts
```diff
@@ -62,6 +64,24 @@ function printService(printer: Printer, service: RPCDescriptor) {
...
+  service.methods
+    .forEach(method => {
+      if (!method.requestStream && !method.responseStream) {
+        printer.printIndentedLn(`${method.nameAsCamelCase}(call: grpc.ServerUnaryCall<${method.requestType}, ${method.responseType}>, callback: grpc.sendUnaryData<${method.responseType}>): void;`);
+      } else if (!method.requestStream) {
+        printer.printIndentedLn(`${method.nameAsCamelCase}(call: grpc.ServerWritableStream<${method.requestType}, ${method.responseType}>): void;`);
+      } else if (!method.responseStream) {
+        printer.printIndentedLn(`${method.nameAsCamelCase}(call: grpc.ServerReadableStream<${method.requestType}, ${method.responseType}>, callback: grpc.sendUnaryData<${method.responseType}>): void;`);
+      } else {
+        printer.printIndentedLn(`${method.nameAsCamelCase}(call: grpc.ServerDuplexStream<${method.requestType}, ${method.responseType}>): void;`);
```
badsyntax, line 79: For better compatibility of types between `grpc` and `@grpc/grpc-js`, us the following method definitions:

- grpc.handleUnaryCall
- grpc.handleServerStreamingCall
- grpc.handleClientStreamingCall
- grpc.handleBidiStreamingCall

IchordeDionysos:
@badsyntax what's the status of this PR? 🤔

badsyntax:
@IchordeDionysos unfortunately i'm still waiting on these changes to be released:

https://github.com/grpc/grpc-node/pull/1587
https://github.com/grpc/grpc-node/pull/1590

i need those changes to build a generic service interface that will work with both `grpc` and `@grpc/grpc-js` packages

badsyntax:
https://github.com/grpc/grpc-node/pull/1587 is published so this PR is good for a review!

badsyntax:
@MarcusLongmuir Any thoughts about this one? It's not a huge change but would be helpful for people i reckon.

package.json
```diff
@@ -33,6 +33,7 @@
     "google-protobuf": "^3.6.1"
   },
   "devDependencies": {
+    "@grpc/grpc-js": "^1.2.0",
```
MarcusLongmuir, line 36: This is probably best expressed in the README rather than this package specifying it as a dev dependency as that might not get noticed.

MarcusLongmuir (approved):
Thanks for the PR. Minor note about the dependency, but otherwise LGTM 👍

package.json
badsyntax: I think adding it as a dev dependency was intended to make it easier to test the generated types, but it's not required by the package so I have remove it. 
I've also added a compatibility note to the README.

badsyntax:
Thanks for the merge @MarcusLongmuir. Is there an ETA for a release?

MarcusLongmuir:
Apologies for the delay. We've just been overhauling a lot of permissions and repo configurations that made this rather laborious.

This has gone out in `0.14.0`: https://www.npmjs.com/package/ts-protoc-gen/v/0.14.0

badsyntax:
Awesome thank you!

PR #248, opened 2020-11-19
set and clear methods return message instance for method chaining

kcirtaptrick:
## Changes

- set* and clear* methods return their parent message instance instead of void

## Verification

TSLint and tests passed, tried generating files in a project and they're working properly.


mohammadavood:
please merge this feature. method chaining is critical and important

MarcusLongmuir:
Hi both. Apologies for the delay on this. This is significant change to the types and needs tests to improve confidence that the types match the runtime behaviour.

Could you please:
* Update the [`generate.js`](../blob/master/generate.js#L6) file to use the latest version of protoc (the currently-used version doesn't have chaining).
* Add tests for chaining and assert the return types and values of the set functions.
* Add chaining to the `add` functions. From the docs and [the protoc write code](https://github.com/protocolbuffers/protobuf/blob/c17e3c134c16f9c39424d2342f723344fb0596dd/src/google/protobuf/compiler/js/js_generator.cc#L2936) the `add` functions for primitive repeated fields also return the instance, but `add` functions for non-primitives appear to return the added value as per the [protobuf reference](https://developers.google.com/protocol-buffers/docs/reference/javascript-generated):
> addFoo(): Appends a value of foo to the end of the list of foos that was in the message. Returns the outer message for chaining only if the added value was a primitive. For added messages, returns the message that was added

kcirtaptrick:
Just made 2 of the changes, how would I go about wriitng tests for this? Would I create a new integration test or add to one of the existing ones?

kcirtaptrick:
Closing this PR since I've since moved to protobuf.js and no longer need this change. If someone else wants to continue this effort, feel free

PR #269, opened 2021-03-24
Update protoc & node

MarcusLongmuir:
Note: A lot of the `@types/google-protobuf` getters now have values as optional. It seems pragmatic to assume they are present due to the usage pattern of this data coming from the protoc compiler that will set them.

## Changes

* Update `protoc`, `google-protobuf`, and `@types/google-protobuf`
* Update node

## Verification

Local and CI run


easyCZ (approved):
/lgtm

src/ExportMap.ts
```diff
@@ -30,12 +30,12 @@ export class ExportMap {
...
-      messageOptions: message.getOptions(),
-      mapFieldOptions: message.getOptions() && message.getOptions().getMapEntry() ? {
-        key: [message.getFieldList()[0].getType(), message.getFieldList()[0].getTypeName().slice(1)],
-        value: [message.getFieldList()[1].getType(), message.getFieldList()[1].getTypeName().slice(1)],
+      pkg: fileDescriptor.getPackage()!,
+      fileName: fileDescriptor.getName()!,
+      messageOptions: message.getOptions()!,
+      mapFieldOptions: message.getOptions() && message.getOptions()!.getMapEntry() ? {
+        key: [message.getFieldList()[0].getType()!, message.getFieldList()[0].getTypeName()!.slice(1)],
+        value: [message.getFieldList()[1].getType()!, message.getFieldList()[1].getTypeName()!.slice(1)],
```
moadz, line 38: I'm assuming the API has changed to allow for `undefined` return types. Are we confident all these nullsafe operators won't bite us? Might be safer reflect the `undefined` type in the `ExportMessageEntry` type.

moadz (approved):

src/ExportMap.ts
MarcusLongmuir: I decided to redo this and added assertions using a `throwError` util function to avoid the explosion of separate checks. Can you give it a quick look to see what you think please?

moadz: LGTM! Thanks for addressing.

PR #272, opened 2021-04-02
Fix get/set conflicting method names.

### LINKED ISSUE #271: Incorrect get/set methods generated for 'extension'/'js_pb_message_id' fields.
There are methods defined on the base protobuf Message class which conflict with the default getter/setter methods. The protoc compiler works around this by adding a '$' suffix to the methods so there are no conflicts, however ts-protoc-gen doesn't use the same algorithm.

Reference: https://github.com/protocolbuffers/protobuf/blob/97cb3a862f132c6ab98e240be11990b89f7d5467/src/google/protobuf/compiler/js/js_generator.cc#L528-L531

**Versions of relevant software used**
0.14.0

**What happened**
Typescript types for field named 'extension' are incorrect. The base message class has methods 'getExtension' and 'setExtension' and so the protoc compiler adds a '$' suffix on the getter/setter method names to not conflict.

**What you expected to happen**
Types to match the generated JS code from protoc.

**How to reproduce it (as minimally and precisely as possible)**:
Create a simple message:

```proto
message GetterNameConflictMessage {
    string extension = 1;
    bytes js_pb_message_id = 2;
}
```

Run the protoc compiler and ts-protoc-gen. Notice that the typescript types have conflicting methods 'getExtension', 'setExtension', 'getJsPbMessageId', 'setJsPbMessageId'

**Full logs to relevant components**

n/a

**Anything else we need to know**

n/a

pkwarren:
Update the typescript generator to add a suffix '$' to methods which
conflict on the Message base class
('getExtension'/'setExtension'/'getJsPbMessageId'/'setJsPbMessageId').
This will keep the typescript definitions in sync with the generated JS
code from the protoc compiler.

Fixes #271.

## Changes

- Added example getter_name.proto which demonstrates conflicts with both string and bytes types.
- Added integration test verifying the generated code.
- Added suffix '$' on getter/setter methods.

## Verification

- Ran 'npm run test' to ensure test passed.
- Verified that the typescript compiler didn't show any warnings about method conflicts in generated *.d.ts files.


examples/generated-grpc-js-node/proto/examplecom/getter_name_pb.js
```diff
@@ -0,0 +1,224 @@
...
+    );
+  }
+};
+
+
+/**
+ * optional string extension = 1;
+ * @return {string}
+ */
+proto.examplecom.GetterNameConflictMessage.prototype.getExtension$ = function() {
```
pkwarren, line 168: Here you can see the protoc compiler adds `$` to get and set functions.

examples/generated-grpc-js-node/proto/examplecom/getter_name_pb.js
```diff
@@ -0,0 +1,224 @@
...
+proto.examplecom.GetterNameConflictMessage.prototype.setExtension$ = function(value) {
+  return jspb.Message.setProto3StringField(this, 1, value);
+};
+
+
+/**
+ * optional bytes js_pb_message_id = 2;
+ * @return {!(string|Uint8Array)}
+ */
+proto.examplecom.GetterNameConflictMessage.prototype.getJsPbMessageId$ = function() {
```
pkwarren, line 186: It does the same for bytes get/set methods.

src/ts/message.ts
```diff
@@ -201,8 +207,8 @@ export function printMessage(fileName: string, exportMap: ExportMap, messageDesc
         }
         const fieldObjectName = normaliseFieldObjectName(camelCaseName);
         toObjectType.printIndentedLn(`${fieldObjectName}${canBeUndefined ? "?" : ""}: ${fieldObjectType},`);
-        printer.printIndentedLn(`get${withUppercase}(): ${exportType}${canBeUndefined ? " | undefined" : ""};`);
-        printer.printIndentedLn(`set${withUppercase}(value${type === MESSAGE_TYPE ? "?" : ""}: ${exportType}): void;`);
+        printer.printIndentedLn(`get${jsGetterName(withUppercase)}(): ${exportType}${canBeUndefined ? " | undefined" : ""};`);
+        printer.printIndentedLn(`set${jsGetterName(withUppercase)}(value${type === MESSAGE_TYPE ? "?" : ""}: ${exportType}): void;`);
```
pkwarren, line 211: I tried to also verify if field presence is enabled, but the TS generation fails. It is possible there might be conflicts with additional methods (hasExtension, etc.) when it is supported.

pkwarren:
@MarcusLongmuir - Is there any additional steps I should follow for PR reviews or is assigning @easyCZ still the correct process?

moadz (approved):
LGTM, but @MarcusLongmuir will also be taking a look

src/ts/message.ts
MarcusLongmuir: I think you were hitting [this issue](https://github.com/improbable-eng/ts-protoc-gen/pull/276) which is now fixed on `master`.

I've added the presence checks in [this commit](https://github.com/improbable-eng/ts-protoc-gen/commit/d4822f87a1d7ba0bda47e8e64ef5a6aeefeddf5a#diff-d0efdb75df2ff73ffd47aefae54c6e34b04e13e1cf6819e34e4da3b2b2b7219aL1). Can you add them to this PR please?

pkwarren: I've rebased to master and pulled in that change - thanks!

src/ts/message.ts
```diff
@@ -157,12 +163,12 @@ export function printMessage(fileName: string, exportMap: ExportMap, messageDesc
     function printClearIfNotPresent() {
       if (!hasClearMethod) {
         hasClearMethod = true;
-        printer.printIndentedLn(`clear${withUppercase}${field.getLabel() === FieldDescriptorProto.Label.LABEL_REPEATED ? "List" : ""}(): void;`);
+        printer.printIndentedLn(`clear${jsGetterName(withUppercase)}${field.getLabel() === FieldDescriptorProto.Label.LABEL_REPEATED ? "List" : ""}(): void;`);
```
pkwarren, line 166: Noticed that the clear method name didn't match the generated code. This is now fixed as well (and tests updated).

MarcusLongmuir (approved):

PR #276, opened 2021-04-21
Fix primitive extension handling

### LINKED ISSUE #273: Non-message extension fields cause 0.14.1 to fail
When testing the latest version (0.14.1) against a .proto file that defines extensions with primitive field types, we're now hitting an error during generation of the typescript types.

**Versions of relevant software used**

0.14.1 (not published to NPM but fetched via github ref)

**What happened**

ts-protoc-gen failed with an error.

**What you expected to happen**

No errors during type generation. The same proto files compile on 0.14.0.

**How to reproduce it (as minimally and precisely as possible)**:

Create an extension that uses non-message field types, i.e.:

```proto
extend google.protobuf.MessageOptions {
  bool my_field = 5000;
}
```

**Full logs to relevant components**

<details>Logs
<p>

```
protoc-gen-ts error: Error: Missing extension type name - This is unhandled behaviour and should be reported as an issue on https://github.com/improbable-eng/ts-protoc-gen/issues
    at Object.throwError (.../node_modules/ts-protoc-gen/lib/util.js:12:15)
    at Object.printExtension (.../node_modules/ts-protoc-gen/lib/ts/extensions.js:11:63)
```

</p>
</details>

**Anything else we need to know**

n/a

MarcusLongmuir:
## Changes

* Getter check prevented `null` being acceptable for the type name which is the case for primitives
* Fixes #273

## Verification

* Added tests for extensions


moadz (approved):
LGTM

PR #277, opened 2021-04-23
Fix CI failing due to diff caused by npm install

MarcusLongmuir:
## Changes

* Stopped caching `node_modules` - appears to have negligible impact on CI duration.
* Fixes CI that [appeared to be](https://travis-ci.org/github/improbable-eng/ts-protoc-gen/builds/768112003#L497) hitting a npm issue that [inexplicably downgrades from https to http when `npm install`ing](https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818).

## Verification

* CI passed for this branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,6 @@ deploy:
   tag: "$NPM_TAG"
   on:
     branch: master
-cache:
-  directories:
-  - node_modules
 install:
 - npm install
 script:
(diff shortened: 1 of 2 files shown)


moadz (approved):
Thanks, npm. This is why we can't have nice things.

Issue #218, opened 2020-01-07
Generated setter declarations have incorrect return type


scisci:
**Versions of relevant software used**

0.12.0

**What happened**

The generated return value for setters is void, but protoc actually returns this so you can chain your setters.
```
export class Thing extends jspb.Message {
  getId(): string;
  setId(value: string): void;
}
```
**What you expected to happen**
```
export class Thing extends jspb.Message {
  getId(): string;
  setId(value: string): this;
}
```
**How to reproduce it (as minimally and precisely as possible)**:

Should behave the same for all messages.

Multiply:
@scisci did you end up fixing this yourself? We're in the same situation after updating grpc-tools, and would be interested in either contributing a fix, or help getting something going.

scisci:
@Multiply, no I haven't had time to fix it. This repo seemed rather inactive last I checked though so  I have a general note of possibly switching to this library: https://github.com/thesayyn/protoc-gen-ts but haven't finished researching.

Multiply:
@scisci I ended up maintaining our own fork as well. It was a rather simple change, luckily.

PR #281, opened 2021-06-21
Change return type of generated setters to the message type

craigatron:
## Changes

- message.ts - changed return type of generated setters to ${messageName}
- ran `npm run generate` to regenerate example proto generated code
- added tests that chain setter calls and verify value on generated object

## Verification

Added unit tests for chained setter methods.

### DIFF
--- a/src/ts/message.ts
+++ b/src/ts/message.ts
@@ -184,12 +184,12 @@ export function printMessage(fileName: string, exportMap: ExportMap, messageDesc
         printer.printIndentedLn(`get${withUppercase}List(): Array<Uint8Array | string>;`);
         printer.printIndentedLn(`get${withUppercase}List_asU8(): Array<Uint8Array>;`);
         printer.printIndentedLn(`get${withUppercase}List_asB64(): Array<string>;`);
-        printer.printIndentedLn(`set${withUppercase}List(value: Array<Uint8Array | string>): void;`);
+        printer.printIndentedLn(`set${withUppercase}List(value: Array<Uint8Array | string>): ${messageName};`);
         printRepeatedAddMethod("Uint8Array | string");
       } else {
         toObjectType.printIndentedLn(`${camelCaseName}List: Array<${exportType}${type === MESSAGE_TYPE ? ".AsObject" : ""}>,`);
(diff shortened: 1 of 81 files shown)


src/ts/message.ts
```diff
@@ -184,12 +184,12 @@ export function printMessage(fileName: string, exportMap: ExportMap, messageDesc
         printer.printIndentedLn(`get${withUppercase}List(): Array<Uint8Array | string>;`);
         printer.printIndentedLn(`get${withUppercase}List_asU8(): Array<Uint8Array>;`);
         printer.printIndentedLn(`get${withUppercase}List_asB64(): Array<string>;`);
-        printer.printIndentedLn(`set${withUppercase}List(value: Array<Uint8Array | string>): void;`);
+        printer.printIndentedLn(`set${withUppercase}List(value: Array<Uint8Array | string>): ${messageName};`);
```
moadz, line 187: From the Protoc docs: 
> addFoo(): Appends a value of foo to the end of the list of foos that was in the message. Returns the outer message for chaining only if the added value was a primitive. For added messages, returns the message that was added (b/138079947).

https://developers.google.com/protocol-buffers/docs/reference/javascript-generated

My understanding is this means the return type of `setInternalChildMessage` for example should be the child message type, and not the parent message type. 

```typescript
        // Wrong 
        setInternalChildMessage(value?: OneOfMessage.InternalChildMessage): OneOfMessage;
        
        // Correct
        setInternalChildMessage(value?: OneOfMessage.InternalChildMessage): OneOfMessage.InternalChildMessage;
```

Can you add some test cases to verify this?

gecko655:
This fix is seriously important for the ts-protoc-gen users (and me).
It seems that fixes and tests are good enough (for me).
Would it be possible for the maintainers to review again?

Note: This PR fixes #218 .

MarcusLongmuir (approved):

chris-codaio:
Just ran into this issue trying to update definitions.  It looks like this PR is approved - any reason it can't be checked in and pushed out in a new version?  Anything I can do to help?

thompsonja:
Ping, is there any reason this can't be merged?

gonzojive:
bump

PR #292, opened 2021-11-01
Add plugin param to only generate gRPC types

tatemz:
## Changes

This change adds the `grpcOnly` parameter to the plugin in order to handle the cases where the user wants to exclusively generate the types for the grpc service only.

Prior to this change, users could not use [`grpc-tools`](https://github.com/grpc/grpc-node/tree/master/packages/grpc-tools), [`grpc-web`](https://github.com/grpc/grpc-web), and `ts-protoc-gen` simultaneously.

After this change, it is possible to let `grpc-web` generate the `.d.ts` files (using the `commonjs+dts` or `typescript` options) for the non-gRPC  generated files, while letting `ts-protoc-gen` singularly manage the generation of the `.d.ts` files for `grpc-node`

## Verification

Tested it with the following [`buf.gen.yml`](https://docs.buf.build/tour/generate-code) configuration:

```yml
version: v1

plugins:
  - name: js
    out: ./__generated/proto/js
    opt:
      - import_style=commonjs,binary

  - name: grpc-node
    path: ./node_modules/.bin/grpc_tools_node_protoc_plugin
    out: ./__generated/proto/js

  - name: grpc-web
    path: ./node_modules/.bin/protoc-gen-grpc-web
    out: ./__generated/proto/js
    opt:
      - import_style=commonjs+dts
      - mode=grpcweb

  - name: ts-node
    path: ~/Development/ts-protoc-gen/bin/protoc-gen-ts
    out: ./__generated/proto/js
    opt:
      - service=grpc-node
      - grpcOnly=true

```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,7 @@ import {CodeGeneratorRequest, CodeGeneratorResponse} from "google-protobuf/googl
 import {FileDescriptorProto} from "google-protobuf/google/protobuf/descriptor_pb";
 import {generateGrpcWebService} from "./service/grpcweb";
 import {generateGrpcNodeService} from "./service/grpcnode";
-import {ServiceParameter} from "./parameters";
+import {ServiceParameter, GrpcOnlyParameter} from "./parameters";
 
 /**
  * This is the ProtoC compiler plugin.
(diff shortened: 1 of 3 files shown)


tatemz:
Dev Note: This feature may not be needed by most, but I found that it is helpful when wanting to generate for `grpc-node` and `grpc-web` simultaneously.  Feel free to suggest any alternative methods for accomplishing this.

I was hoping that `grpc-tools` allowed for the `commonjs+dts` option, but it does not, and the `proto-loader` is not a protoc plugin  ((link)[https://github.com/grpc/proposal/pull/183#discussion_r453237722])

PR #294, opened 2022-01-13
fix: don't pass callback as metadata when omitted

nathanfranke:
## Changes

If `callback` is passed but not `metadata`, reset `metadata`.

## Verification

I changed this code manually on my generated files and it works exactly as expected.


nathanfranke:
Still desired

